PR #16, opened 2018-01-08
Change `HasCallback` to `HasCallStack` in the docs

jkachmar:
I think you meant `HasCallStack` rather than `HasCallback`.


snoyberg:
Thanks!

PR #24, opened 2018-01-17
Raw implementation of RIO

snoyberg:
Not really sure this is necessary, but it's always fun to write these :)
Motivating reason: @nh2 discovered recently that StateT and ReaderT were
missing some optimizations leading to `forever` being painfully slow.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -20,6 +20,7 @@ dependencies:
 - exceptions # FIXME what do we do with MonadThrow?
 - filepath
 - hashable
+- ghc-prim
 - microlens
 - mtl
 - text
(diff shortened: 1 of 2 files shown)


NorfairKing:
Have you benchmarked this against the original implementation? (Only half trolling)

snoyberg:
No, I haven't. For the record: we should only consider merging this ridiculous PR if a benchmark proves that it's the fastest implementation.

mgsloan:
Did this get benchmarked?  I suppose we can hopefully trust GHC to optimize to the equivalent.  To me it seems good to be as fast as possible for these operations that get used over and over.

Issue #17, opened 2018-01-08
Rewrite export statements for nicer Haddocks


snoyberg:
Currently, we get basically no exposed documentation. We need a rewrite that looks like this:

```diff
diff --git a/src/RIO/Prelude.hs b/src/RIO/Prelude.hs
index 09accaf..c8f6b91 100644
--- a/src/RIO/Prelude.hs
+++ b/src/RIO/Prelude.hs
@@ -39,11 +39,18 @@ module RIO.Prelude
   , writeFileDisplayBuilder
   , hPutBuilder
   , sappend
+  , Control.Applicative.Alternative
+  , Control.Applicative.Applicative (..)
+  , Control.Applicative.liftA
+  , Control.Applicative.liftA2
+  , Control.Applicative.liftA3
+  , Control.Applicative.many
+  , Control.Applicative.optional
+  , Control.Applicative.some
+  , (Control.Applicative.<|>)
   ) where
 
-import           Control.Applicative  as X (Alternative, Applicative (..),
-                                            liftA, liftA2, liftA3, many,
-                                            optional, some, (<|>))
+import qualified Control.Applicative
 import           Control.Arrow        as X (first, second, (&&&), (***))
 import           Control.DeepSeq      as X (NFData (..), force, ($!!))
 import           Control.Monad        as X (Monad (..), MonadPlus (..), filterM,
```

PR #26, opened 2018-01-18
Improve haddock generation (see #17)

adituv:
Rewrite export list for RIO.Prelude such that all reexports other than those from UnliftIO are rendered on the RIO haddock page.  As UnliftIO is such a huge import I left that as a module link pending what you want to do with it.

Rewrite export lists in all other modules that only reexport definitions such that their haddocks have links to the source modules.  This probably should be improved later but in my opinion will do for now as a quick fix.

Addresses issue #17

### DIFF
--- a/src/RIO.hs
+++ b/src/RIO.hs
@@ -1,4 +1,5 @@
-module RIO (module X) where
+module RIO ( module RIO.Logger
+           , module RIO.Prelude) where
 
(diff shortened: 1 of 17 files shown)


snoyberg:
This looks great, thank you! If you have other ideas for improving the output here, please don't hesitate to bring it up.

Issue #35, opened 2018-01-26
Namespace conflicts between RIO and RIO.Map


tejon:
`RIO` exports `Data.Foldable.null`. `RIO.Map` (implicitly) exports `Data.Map.Strict.null`. Being not technically the same function (despite the former being defined as the latter), they conflict. There are quite a few other examples of this (e.g. most of `Foldable`) -- `null` is just the first one I ran into.

This comes down to base `Data.Map` being intended for qualified import, so they didn't use unique names; whereas Rio doesn't suggest a default qualifier, so I assume it's intended for use without one. On the upside, the fix should be simply hiding the conflicts in `RIO.Map`'s import.

snoyberg:
The only reason it didn't recommend a qualified name is that I having
written that documentation yet, but it's definitely intended to be
qualified. PR welcome if you'd like to add the missing comment.

tejon:
D'oh! What I get for assuming. I can take care of that. Does this apply to everything that would be qualified under standard Prelude? At the moment it looks like only `RIO.ByteString` has a canonical qualifier.

snoyberg:
PRs are definitely welcome. However, for the moment, there's a bit of a debate about the correct naming, see #12. I'd say that, for now, _any_ kind of recommended naming is fine, and when #12 is resolved we can update the recommendations.

snoyberg:
Addressed by #36, closing.

Issue #12, opened 2018-01-03
Qualified imports proposal


chrisdone:
I propose we take the braindead simple approach of:

```haskell
import Rio
import qualified Rio.Exe as Exe
import qualified Rio.List as List
import qualified Rio.Vector as Vector
import qualified Rio.Text as Text
```
(The rule being `Rio.<foo>` is qualified as `<foo>`).

Where `Rio` exports in the namespace the names of types, and generic functions like `fmap` and `(<>)`, but functions/values are qualified. Lazy versions of things can have an alias like `LazyText` or unboxed things like `UnboxedVector`.

I can add support to Intero that will for example just automatically import `import qualified Rio.Text as Text` to your imports list if you write e.g.

``` haskell
main = writeUtf8 (Text.intercalate ", " xs)
```

So while the list grows, if it's automatically managed, it comes for free. So you wouldn't actually write your own import lists, the editor would do it.

snoyberg:
Personally, I'd rather have the shorter qualified names we usually use today, e.g. `B` for `ByteString`, `BL` for lazy `ByteString`, `T` for `Text`, etc.

Also: I've been using `RIO` upper case for module names; is that a problem?

NorfairKing:
I also prefer:

```
import qualified Data.ByteString as SB
import qualified Data.ByteString.Char8 as SB8
import qualified Data.ByteString.Lazy as LB
import qualified Data.ByteString.Lazy.Char8 as LB8
```

chrisdone:
> Also: I've been using RIO upper case for module names; is that a problem?

From a tooling perspective it shouldn't matter; I'll just be typing lowercase from editor support anyway.

> Personally, I'd rather have the shorter qualified names we usually use today, e.g. B for ByteString, BL for lazy ByteString, T for Text, etc.

I think that whole `B`/`L` thing is a bit incomplete and everybody (just look at @NorfairKing's --  and elsewhere within our own team) just has slightly different versions of everything and muddles along. It trips me up when people use different things.  Here are some things that I personally regularly find conflicting or otherwise I just randomly pick something.

```haskell
Data.Map  (strict/lazy), Data.HashMap.Strict, Data.IntMap -- as M
Data.Text, Data.Tree, Data.Trie -- as T
Data.Text.Lazy as LT?
Data.Vector, Data.Vector.Unboxed as V, UV, SV, MV? UMV? etc.?
[...]

PR #36, opened 2018-01-29
Pervasive qualified import name documentation

tejon:
Addressing #35.

For now, I went with a sort of middle ground with regards to current discussion in #12: the first layer of modules (`RIO.*`) have single-letter abbreviations, but their children are unabbreviated, e.g.:
```
import qualified RIO.Vector.Unboxed as V.Unboxed
```

The exception to this is `RIO.HashMap` and `RIO.HashSet`, which are `M.Hash` and `S.Hash` respectively, reflecting a vague hunch that it wouldn't hurt to move them in the hierarchy accordingly.

### DIFF
--- a/src/RIO/ByteString/Lazy.hs
+++ b/src/RIO/ByteString/Lazy.hs
@@ -1,3 +1,6 @@
+-- | Strict @ByteString@. Import as:
+--
+-- > import qualified RIO.ByteString.Lazy as B.Lazy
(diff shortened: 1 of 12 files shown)


tejon:
...and of course I totally missed a spot in every copypasta. Fixing that..

snoyberg:
Thanks!

PR #50, opened 2018-02-23
README: add app-wide exception type suggestion

kerscher:


README.md
```diff
@@ -327,6 +327,25 @@ basic idea is:
 * Do all resource allocations with functions like `bracket` and
   `finally`.
 
+It’s a good idea to define an app-wide exception type:
+
+```haskell
+data AppExceptions
+  = NetworkChangeError
```
snoyberg: My only concern is that, by providing only constructor with no data, we're implying that's the way an app-wide exception type should be constructed.

kerscher: Would something like this be a better example?

```haskell
data AppExceptions
  = NetworkChangeError Text
  | FilePathError FilePath
  | ImpossibleError
  deriving (Typeable)

instance Exception AppExceptions

instance Show AppExceptions where
  show =
    \case
      NetworkChangeError err -> "network error: " <> (unpack err)
      FilePathError fp -> "error accessing filepath at: " <> fp
      ImpossibleError -> "this codepath should never have been executed. Please report a bug."
```

snoyberg: Yes, I think that gets the idea across nicely.

PR #53, opened 2018-03-05
Remove partial functions from reexports

### LINKED ISSUE #43: Remove partial functions from reexports
A number of modules we reexport provide partial functions. We should remove those, or at least put them in some kind of partial function modules.

adituv:
This removes partial and unsafe functions from reexport modules, and
moves them to their own modules.  As a side-effect of this,
documentation in many modules has been improved as there are now
explicit export lists.

I have not touched RIO.Process as I am entirely unfamiliar with the API.
To the best of my knowledge, RIO.Directory, RIO.FilePath, and RIO.Time
do not contain any partial functions so are also not touched.

There are various CPP conditions in the export lists for Set, Map,
and the Vector modules (possibly among others) for compatibility with
lts-7.  Let me know if you would like me to change this.

Resolves #43 

Edit: Apologies for the upload spam.  I had some problems with different exposed API in `containers` and `vector` over the different versions in other lts's.

### DIFF
--- a/src/RIO/ByteString/Lazy.hs
+++ b/src/RIO/ByteString/Lazy.hs
@@ -2,7 +2,140 @@
 --
 -- > import qualified RIO.ByteString.Lazy as B.Lazy
 module RIO.ByteString.Lazy
-  ( module Data.ByteString.Lazy
+  (
+  -- * The @ByteString@ type
+    Data.ByteString.Lazy.ByteString
+
+  -- * Introducing and eliminating 'ByteString's
+  , Data.ByteString.Lazy.empty
+  , Data.ByteString.Lazy.singleton
+  , Data.ByteString.Lazy.pack
+  , Data.ByteString.Lazy.unpack
+  , Data.ByteString.Lazy.fromStrict
+  , Data.ByteString.Lazy.toStrict
+  , Data.ByteString.Lazy.fromChunks
+  , Data.ByteString.Lazy.toChunks
+  , Data.ByteString.Lazy.foldrChunks
+  , Data.ByteString.Lazy.foldlChunks
+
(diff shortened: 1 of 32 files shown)


src/RIO/ByteString/Lazy.hs
```diff
@@ -2,7 +2,140 @@
...
+  , Data.ByteString.Lazy.hGetContents
+  , Data.ByteString.Lazy.hGet
+  , Data.ByteString.Lazy.hGetNonBlocking
+  , Data.ByteString.Lazy.hPut
+  , Data.ByteString.Lazy.hPutNonBlocking
+  , Data.ByteString.Lazy.hPutStr
   ) where
 
-import Data.ByteString.Lazy
+-- TODO lifted versions of the above I/O functions
```
snoyberg: That's a good catch, we should probably try to include that in the first stable release too.

src/RIO/Set/Unchecked.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE CPP #-}
+
+-- | This module contains functions from "Data.Set" that have unchecked
```
snoyberg: Very nice catch!

src/RIO/Text.hs
```diff
@@ -1,20 +1,151 @@
+{-# LANGUAGE MagicHash #-}
```
snoyberg: I'm surprised this is necessary.

adituv: Me too, but exporting `unpackCString#` without -XMagicHash is a syntax error :/

snoyberg: Doh, I missed that one!

PR #60, opened 2018-03-10
Reexport Data.Bool.bool

tejon:
Noticed this was missing. There may be a principled reason for that, if so disregard.


adituv:
+1 from me.  I suspect it was just an oversight from when I was adding explicit export lists.

PR #62, opened 2018-03-12
Reexport toAscList, toDescList from RIO.Map.Unchecked

tejon:
Looks like these got missed. I put them in `Unchecked` because the matching `from*` are unchecked, and (I believe?) `toAscList . fromAscList = id` so you can technically get an unsorted list out of it. Arguably this is covered by `from*` being in `Unchecked`, in which case  these can move to `RIO.Map` proper; I'm on the fence about it.

### DIFF
--- a/src/RIO/Map/Unchecked.hs
+++ b/src/RIO/Map/Unchecked.hs
@@ -12,11 +12,13 @@ module RIO.Map.Unchecked
 
   -- * Conversion
   -- ** Ordered lists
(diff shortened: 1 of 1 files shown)


adituv:
Oops, these were definitely in RIO.Map at one point but I must have accidentally removed them from the export list when I moved the from counterparts!  They were not ever in unchecked as they do not invalidate internal invariants, so i am indeed in favour of putting them in `RIO.Map` proper instead.

PR #63, opened 2018-03-12
Reexport Data.Monoid.Alt from Prelude

tejon:
Not actually searching for these, I just keep reaching for the one thing that fell of the shelf!


snoyberg:
Thanks!

PR #65, opened 2018-03-13
Add the rio-orphans package

snoyberg:
Also moves us over to a megarepo setup


tejon:
> Also moves us over to a megarepo setup

Off topic, but is there a way I can still use the repository in `extra-deps:` ? It's choking on not having a top-level .cabal file.

snoyberg:
Yes, via subdirs. Check out
https://github.com/snoyberg/mega-sdist/blob/master/stack.yaml

PR #67, opened 2018-03-15
Add more exports from UnliftIO.Concurrent

roman:
I think it is useful to have some more exports from `UnliftIO.Concurrent`. 

Another option would be to re-export `UnliftIO.Concurrent` as `RIO.Concurrent`, let me know your thoughts

### DIFF
--- a/rio/src/RIO/Prelude/Reexports.hs
+++ b/rio/src/RIO/Prelude/Reexports.hs
@@ -2,6 +2,12 @@
 module RIO.Prelude.Reexports
   ( module UnliftIO
   -- List imports from UnliftIO?
+  , UnliftIO.Concurrent.ThreadId
+  , UnliftIO.Concurrent.myThreadId
+  , UnliftIO.Concurrent.isCurrentThreadBound
+  , UnliftIO.Concurrent.yield
+  , UnliftIO.Concurrent.threadWaitRead
+  , UnliftIO.Concurrent.threadWaitWrite
   , UnliftIO.Concurrent.threadDelay
   , Control.Applicative.Alternative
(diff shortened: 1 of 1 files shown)


rio/src/RIO/Prelude/Reexports.hs
```diff
@@ -2,6 +2,12 @@
 module RIO.Prelude.Reexports
   ( module UnliftIO
   -- List imports from UnliftIO?
+  , UnliftIO.Concurrent.ThreadId
+  , UnliftIO.Concurrent.myThreadId
+  , UnliftIO.Concurrent.isCurrentThreadBound
+  , UnliftIO.Concurrent.yield
```
snoyberg: This reexport in particular could be problematic, as it would conflict with streaming libraries (at least conduit). The others have quite unique namings.

roman: I see, maybe we can rename this to `yieldThread`?

snoyberg: Sounds perfect!

rio/src/RIO/Prelude/Reexports.hs
```diff
@@ -299,3 +305,6 @@ import qualified GHC.Stack
 import qualified Prelude
 import qualified System.Exit
 import qualified Text.Read
+
+yieldThread :: MonadIO m => m ()
+yieldThread = UnliftIO.Concurrent.yield
```
tejon: Should this be inlined?

roman: Good catch, I added the inline pragma now

PR #69, opened 2018-03-19
31 trace functions

snoyberg:
Just pulling in the commits from @adituv


tejon:
Re: comments on `trace*IO` -- after thinking about it a bit, I'm in favor of keeping the warning. It's a matter of intent.

snoyberg:
Awesome, thanks for the code and the review!

Issue #71, opened 2018-03-20
Move / expose some RIO.Prelude.* modules


mgsloan:
In PR #72, I propose to not export all the modules under `RIO.Prelude.*`.  However, in some cases you might not want to import `RIO`, yet you still want to use some stuff from it without import lists.  I'm thinking the following are good candidates for renaming and inclusion in `exposed-modules`:

* `RIO.Prelude.Display` -> `RIO.Display`
* `RIO.Prelude.Logger` -> `RIO.Logger`
* `RIO.Prelude.RIO` -> `RIO.Monad`?  Iffy.  `RIO.RIO` seems a little overexcited about RIO.
* `RIO.Prelude.Trace` -> `RIO.Trace` 
* `RIO.Prelude.URef` -> `RIO.URef`?  Also a bit iffy since there isn't a `RIO.IORef`, for example.

So, this would leave the following modules under `RIO.Prelude.___`, and not included in `exposed-modules`: `Extra`, `IO`, `Lens`, `Reexports`, `Renames`, and `Text`.

snoyberg:
This makes the module structure confusing. Until now, if something is in `RIO.*`, it's _not_ exposed by `RIO` itself. This will change around that assumption.

snoyberg:
I thought about this more. Doing the move should be fine, we just need to include a comment in the module header "this is exported by `RIO`" or similar.

jkachmar:
It looks as if this issue was addressed in b23f9c5, can it be closed?

ketzacoatl:
I think so, but @snoyberg can re-open if there is more to do.

snoyberg:
Yeah, let's close, thanks @ketzacoatl.

Issue #61, opened 2018-03-12
More informative documentation for RIO module


mgsloan:
From a documentation perspective, it is quite hard for a user to tell what the `RIO` module exports.  This is because it uses `module` exports.  The re-export depth can be quite high.  For example, `RIO` exports `module RIO.Prelude.Reexports`, and it exports `module UnliftIO`, which itself consists of module exports.  From the perspective of the compiler or heavy ghci user that just uses `:bro RIO` (`:browse`), this doesn't obscure anything at all.  However, for a reader of the documentation, it can be hard to tell what `RIO` really provides.

I started improving the readability of the haddocks for `RIO.Prelude.Reexports` - https://gist.github.com/mgsloan/bec603328664ea7d20075fdddefad5a0 , but I started wondering if it would be better move it into the `RIO` module and switch everything over to explicit exports.  There are some downsides to this idea, though.  One of them is that it may be easy to add something and forget to add it to the re-exports.

So, it seems like the best option would be to improve haddock to allow some annotation on module re exports which causes them to be expanded.

Now, a fair question would be whether we really want that - the full docs for the RIO module might be overwhelming particularly with the large instance lists for many of the classes.  I think it'd be more helpful than harmful, though.  To me it makes sense to be able to read the full docs for your prelude in one page, and also be able to use browser text search within that page.

adituv:
Honestly I am considering whether `RIO.Prelude.Reexports` should be removed entirely and its contents distributed across other modules.  Whether or not something is just reexported from another package isn't really that important for an end-user to know, and having to search multiple places in the documentation for a function definition because you don't know whether it's a reexport is not great.

If we went this route, I think an organisation similar to `base` would work well - reexports should go in modules like `RIO.Text` etc already do, and then the most useful ones could then be exported from `RIO` as well.  This would probably be quite a lot of work though.

As for improving haddock, it already expands exported module documentation for modules that are not exported, so for everything other than `UnliftIO` we could just add them to `other-modules`.

mgsloan:
@adituv Agreed, Reexports is a bit weird, seems like it's mostly organization for development of RIO.  PR #72 hides export of those modules.  It's rather unorganized, though, and results in a funny mishmash of section headers:

![2018-03-20_395x667_scrot](https://user-images.githubusercontent.com/22570/37644752-03e4fba6-2be2-11e8-9d12-3955c8808925.png)

lehins:
[...]

PR #72, opened 2018-03-20
Hide RIO.Prelude.* modules

mgsloan:
Many of these modules seem a bit out of place, or otherwise only make sense in
the context of being re-exported in Rio. Particularly, the following only make
sense in that context: Extra, IO, Reexports, Renames, Text.

As such, I think for the stable release it makes sense to just hide all of them.
In later releases, additional modules / exports can be added to allow importing
of some of these things individually. This is described in #71

### DIFF
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -45,6 +45,18 @@ when:
 
 library:
   source-dirs: src/
+  other-modules:
+    - RIO.Prelude.Display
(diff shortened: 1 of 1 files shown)


mgsloan:
This is also a bit of a step towards #61, as nearly all of the exports of `RIO` end up in the module documentation.  Unfortunately, the docs are rather disorganized.

PR #73, opened 2018-03-20
Remove stack-specific unicode logic from RIO.Prelude.Logger

mgsloan:
It seems like this should be done differently in stack anyway, directly munging particular process output instead of applying to all of stack's logging.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -258,8 +258,8 @@ canUseUtf8 h = liftIO $ wantWritableHandle "canUseUtf8" h $ \h_ -> do
 -- return both a 'LogOptions' value and an 'IORef' containing the
 -- resulting 'Builder' value.
 --
--- This will default to non-verbose settings and assume there is a
--- terminal attached. These assumptions can be overridden using the
+-- This will default to non-verbose settings, use utf8, and assume there
+-- is a terminal attached. These assumptions can be overridden using the
 -- appropriate @set@ functions.
 --
 -- @since 0.0.0.0
@@ -272,6 +272,7 @@ logOptionsMemory = do
         , logTerminal = True
         , logUseTime = False
         , logUseColor = False
+        , logUseUtf8 = True
         , logSend = \new -> atomicModifyIORef' ref $ \old -> (old <> new, ())
         }
   return (ref, options)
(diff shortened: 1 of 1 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -287,41 +288,18 @@ logOptionsHandle
...
-          else do
-            let lbs = toLazyByteString builder
-                bs = toStrictBytes lbs
-            case decodeUtf8' bs of
-              Left e -> error $ "mkLogOptions: invalid UTF8 sequence: " ++ show (e, bs)
-              Right text -> do
-                let text'
-                      | unicode = text
-                      | otherwise = T.map replaceUnicode text
-                TIO.hPutStr handle' text'
```
snoyberg: This branch is definitely _not_ Stack specific. In the new version of the code, you seem to be ignoring the `logUseUtf8` value entirely. This code will not work on any machine with a locale/code page setting that does not support UTF-8.

mgsloan: If you take a look at the replaceUnicode function, it only replaces two codepoints specific to GHC output.

Yes, `logUseUtf8` is ignored entirely, it's intended to be read by the user.  On further review, that is probably unexpected and probably should have an accompanying `askUseUtf8` or something.

So, what should we do here?  There are a few different ways to munge unicode into 7 bit ASCII

mgsloan: Do other loggers just punt on this?  Why is it particularly coming up here?  Just that stack has been used a bunch on a variety of platforms?

snoyberg: I think you're missing the exact line of code that I'm commenting on. The way the code works today is that, if the file handle doesn't support UTF8, we decode the bytestring to Text and then use the textual I/O functions to output it.

mgsloan: I see.  I'm not familiar with how `hPutBuilder handle' (builder <> flush)` would differ from `T.hPutStrLn (decodeUtf8 ...)`.  Seems like it should be pretty equivalent.

mgsloan: Ah, I guess it's because in theory it will work for things that aren't utf-8.  Having a hard time finding where in the code that happens, though.  I guess the stuff in System.IO does that.  It looks like the text io stuff will have different concurrent behavior, potentially interspersing output of chars.

snoyberg:
Discussed offline. The code is _not_ Stack specific, and needs to remain. There's an argument to be made around the quote replacement, but overall it does good without major harm (just a single traversal of a `Text`), so leaving.

PR #74, opened 2018-03-20
Remove stripCR function

mgsloan:
The main place this is useful is after use of some 'lines' function that splits
only on \n.  I think for these cases it makes more sense to have new 'lines'
functions which handle both \r\n and \n.

### DIFF
--- a/rio/src/RIO/List.hs
+++ b/rio/src/RIO/List.hs
@@ -135,6 +135,7 @@ module RIO.List
 
   -- ** Functions on strings
   , Data.List.lines
+  , linesCR
   , Data.List.words
   , Data.List.unlines
   , Data.List.unwords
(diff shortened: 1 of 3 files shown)


snoyberg:
What about adding such new `lines` functions at the same time? Note that we should use a different name to make the difference in behavior clear.

mgsloan:
Makes sense.  Right now I'm having trouble thinking of a good name, though.  I suppose could always throw a `'` on it.

snoyberg:
How about `linesCR`? It's clunky, but it gets the difference across _really_ clearly. And for someone who doesn't know what CR stands for, it's just begging them to look at the docs.

mgsloan:
Good idea! I've added a commit which adds those.

Issue #33, opened 2018-01-21
RIO's (<>) conflicts with with Data.Semigroup's


tejon:
I.e., you can't declare a Semigroup instance without hiding `RIO.(<>)`. Not sure how much of an issue this is in practice, but it doesn't seem like it should have to be the case?

snoyberg:
I'm classy-prelude we used <> for semigroup and ++ for Monoid. It works
well in practice IMO, but not sure if that's an acceptable change from
standard practice for rio. On the other hand, with 8.4 coming out soon,
most of the missing instances for Semigroup are being added, so it may not
make much of a difference.

snoyberg:
I pushed through this change in https://github.com/commercialhaskell/rio/commit/cb1cdedac2edce9c19619b433b2709bbc7e0350c#diff-fc787bce63e5df57e918a279f69f25e1L175. If there are still issues with this, please comment, otherwise closing for now.

snoyberg:
Actually, I'm going to revert this and reopen the issue for discussion. Trying out this change in the Stack codebase leads to too much friction with existing libraries that reexport `<>` and `++` from base.

tejon:
Is the friction you mentioned with `<>`, or only with `++`? I'm not sure we need the latter, seeing as Rio is not supporting older versions of base prior to Semigroup becoming a superclass of Monoid. `Semigroup.<>` should always work for both, no?

Granted, I'm completely baffled as to how and why they aren't the same function to begin with, and the answer here probably depends on that. A brief hackage source check tells me nothing; they both point to the same definition in `GHC.Base`.

snoyberg:
>  seeing as Rio is not supporting versions of base prior to Semigroup becoming a superclass of Monoid.

That's not true. We're supporting from GHC 8.0. I think that's where a lot of the misunderstanding going on here is coming from.

snoyberg:
I've opened PR #80 which switches to Semigroup's `<>` usage. This may cause some friction short term, as libraries are still updating to the Semigroup/Monoid switch, but given that 8.4.1 is out the door, that friction should hopefully disappear quickly. To test this out, I've updated Stack's rio branch to use this change:

https://github.com/commercialhaskell/stack/commit/3daafdc2b502f64fb0935bef98a3dcb936bc8554

I think this addresses the original issue here. If so, I think we can merge this and then make a first release of rio.

PR #75, opened 2018-03-20
Remove some Data.Monoid wrappers

mgsloan:
Reasoning:

* Kept 'First' / 'Last' because they are useful in many practical circumstances.
  Particularly when used with fields of a datatype + Monoid deriving.

* 'Alt' is a very general name, and off the top of my head I didn't know what it
  does.

* 'Product' / 'Sum' are pretty clear, but they are also rather popular type
  constructor names, if you search them in hoogle.  A lot of collisions and not
  used a whole ton.

* 'All' / 'Any' are also pretty clear, but similarly a fairly popular type name.

* Kept 'Endo' because I like it, though tbh I don't use it much.

### DIFF
--- a/rio/src/RIO/Prelude/Reexports.hs
+++ b/rio/src/RIO/Prelude/Reexports.hs
@@ -149,15 +149,7 @@ module RIO.Prelude.Reexports
   , Data.Maybe.mapMaybe
   , Data.Maybe.maybe
   , Data.Maybe.maybeToList
-  , Data.Monoid.All (..)
-  , Data.Monoid.Any (..)
-  , Data.Monoid.Endo (..)
-  , Data.Monoid.First (..)
-  , Data.Monoid.Last (..)
-  , Data.Monoid.Alt (..)
   , Data.Monoid.Monoid (..)
-  , Data.Monoid.Product (..)
-  , Data.Monoid.Sum (..)
   , (Data.Monoid.<>)
   , Data.Ord.Ord(..)
   , Data.Ord.Ordering(..)


tejon:
I _just_ added Alt last week when I discovered (a) it was the one thing not exported and (b) it was exactly what I needed to preserve a list of keys to be queried when the return from the database might be null for some of them (and thus return nothing, losing the knowledge that the key was asked for).
```Haskell
mergeSkeleton ids = fmap getAlt . M.unionWith (<|>) (mkSkeleton ids) . fmap Alt
mkSkeleton = M.fromList . fmap (id &&& const mempty) . toList
```

It is also a generalization of `First` and `Last`, as explained in their documentation.

That said, I'm not entirely against moving _all_ of `Data.Monoid` to `RIO.Monoid`, possibly adding `RIO.Semigroup`, and exporting only `Semigroup.(<>)`, `mappend`, `mempty` and `mconcat` from `RIO` proper.

mgsloan:
Yes, I definitely prefer (<>) from Semigroup.  Wouldn't mind having `RIO` only export `Monoid` and none of the newtypes.

snoyberg:
In line with my general rule of "we should export things if they have unambiguous names," I'll make an argument in favor of not putting these in `RIO` itself: IIRC, the semigroups package has alternate definitions of many of these newtypes. I'd be totally in favor of a `RIO.Monoid` vs `RIO.Semigroup` module, or something along those lines.

mgsloan:
I've updated the PR to remove all the Data.Monoid newtypes.

Perhaps we should also leave out `(<>)`?  Downside is that then people need to import either `RIO.Monoid` or `RIO.Semigroup` once those exist.

tejon:
IMO just add those two modules as part of this PR.

tejon:
...oh, that doesn't answer the question of whether to reexport. I think yes, from `Data.Semigroup`.

Actually, glancing at the sources right now, I'm thinking maybe we only need `RIO.Semigroup` at all, since the only things actually defined in `Data.Monoid` are `First` and `Last`, which as mentioned earlier are generalized by the newer `Alt`. The other newtypes exposed in `Data.Monoid` are originally from `Data.Semigroup.Internal`!

mgsloan:
Yeah, I suppose we could add those modules with this PR.  I'm leaning towards removing things before the stable release rather than adding things hastily, though.

snoyberg:
@tejon I'm not following the discussion that terribly well, but are you OK with me merging this?

tejon:
If wet doing this separate from adding `RIO.Semigroup`/`RIO.Monoid`, then I think for now the export of `(<>)` should either also be removed, or come from `Data.Semigroup` (preferably the latter if that doesn't cause any other problems).

Reasoning: I discovered a while back that the status quo interferes with declaring a `Semigroup` instance. Previously that was offset by never needing to import `Data.Monoid`.

snoyberg:
I'm not able to understand the connection between what we do with `<>` and the changes here. I'm OK with trying to figure out a solution before the first release for `<>` (though TBH I'm not particular hopeful that a good solution exists), but not sure that it should happen here, interleaved with other discussions.

tejon:
Fair enough -- the connection in my mind is simply that, other than the class itself, `(<>)` is the only remaining export from `Data.Monoid`. Putting that aside, I have no objection here.

snoyberg:
OK, should we continue the `<>` discussion back on #33?

Issue #76, opened 2018-03-20
Should the whole `UnliftIO` module be re-exported?


mgsloan:
My review of the re-exports that come from `UnliftIO`:

* `Control.Monad.IO.Unlift` seems good :+1:

* `UnliftIO.Async` might be ok to re-export.  It's the current defacto standard, sure, but some functions like `wait` / `poll` / `link` seem likely to accidentally alias other things.  Has quite a lot of functions that don't get a ton of use but maybe that's ok.

* `UnliftIO.Chan` does not seem like a good thing to re-export (see relevant section of https://lorepub.com/post/2016-12-17-Haskell-Pitfalls - I googled around a little for a concrete reason backing why I felt like channels weren't so great)

* `UnliftIO.Exception` seems alright :+1: .  Might consider leaving out `impureThrow` / `assert`, but they're probably ok considering that `error` is also exported by RIO.

* `UnliftIO.IORef` :+1:

* `UnliftIO.MVar` :+1:

* `UnliftIO.STM` is mostly good, but it might make sense to omit `TChan` / `TQueue`. Why? They don't seem to get used a whole ton, seems prudent to remove exports when they aren't likely to be used much.  Similar pitfall for channels applies to all but the TBQueue.

* `UnliftIO.Temporary` :+1:  Wish we could use `path`, though!

* `UnliftIO.Timeout` :+1:

snoyberg:
I don't see a reason to avoid reexporting namespaced functions and types, like those associated with `TChan`, `TQueue`, etc. What's the downside to having them easily available?

snoyberg:
FWIW, on reviewing that blog post linked above: I thought about including some bounded, closable channel and queue types ala stm-chans in `unliftio`. I could get behind the idea that we should focus on those instead, and treat the unbounded variants we have now as a mistake like partial functions.

snoyberg:
We've agreed to move the discussion towards fixing exports from `UnliftIO` instead.

PR #78, opened 2018-03-23
Prune some UnliftIO re-exports

mgsloan:
See issue #76


snoyberg:
Travis is failing due to missing dependency. In addition: I'm not sure how good an idea it is to change things like this, as opposed to changing the exports in `UnliftIO` itself.

mgsloan:
Hmm, yeah that's an interesting option.  It would be cleaner if the `Rio` modules was a superset of `UnliftIO`

snoyberg:
We've agreed to move this to the `unliftio` repo instead.

PR #79, opened 2018-03-28
Export safe *May functions for partials in Foldable

tejon:


snoyberg:
These functions are partial, I don't think we should include them.

tejon:
Ah! I completely overlooked that. Agreed.

snoyberg:
I thought about this some more in the car. I'd be totally in favor of
adding functions like `maximumMay` or `maximumMaybe`.

tejon:
In that case I think we should re-export `*May` from [Safe.Foldable](https://hackage.haskell.org/package/safe-0.3.17/docs/Safe-Foldable.html).

snoyberg:
For such short functions, I'd rather not add an extra dependencies.

tejon:
I'd usually agree but the dependency is equally short (zero transients) and has a respectable list of reverse dependencies. I think it's worth avoiding any conflict if someone imports it for all the more special-case functions.

snoyberg:
That logic will eventually lead to dozens of additional dependencies, each
for a small number of functions.

tejon:
I feel like this is a pretty special case, but you may be right. Certainly not the hill I want to die on. :) Closing this for the moment, I don't think I'll have time to get back to it until after the weekend.

PR #84, opened 2018-04-16
Add a template

snoyberg:
This is intended for usage with `stack new`. The code here is certainly up for improvement.

### DIFF
--- a/etc/make-template.hs
+++ b/etc/make-template.hs
@@ -15,6 +15,8 @@ import qualified RIO.ByteString.Lazy as BL
 import RIO.FilePath
 import RIO.Process
 import qualified RIO.Text as T
+import qualified RIO.Text.Partial as T (replace)
+import RIO.Time
 import Text.ProjectTemplate
 
 data App = App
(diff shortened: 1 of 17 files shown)


etc/template/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Author name here (c) 2018
```
mgsloan: Should probably be `AUTHOR` not `Author name here`

etc/template/package.yaml
```diff
@@ -0,0 +1,51 @@
...
+# category:            Web
+
+# To avoid duplicated efforts in documentation and dealing with the
+# complications of embedding Haddock markup inside cabal files, it is
+# common to point users to the README.md file.
+description:         DESCRIPTION
+
+dependencies:
+- base >= 4.10 && < 5
+- rio >= 0.1.1.0 && < 0.2
```
roman: Wondering why the upper-bound here?

snoyberg: Good point, I've removed it.

etc/make-template.hs
```diff
@@ -0,0 +1,81 @@
...
+
+  -- Make sure it passes
+  proc "stack" ["test"] runProcess_
+
+  rawout <- withWorkingDir templatedir
+          $ proc "git" ["ls-files", "-z"]
+            readProcessStdout_
+  files <- mapM decode $ splitNulls $ BL.toStrict rawout
+  let src = forM_ files $ \fp -> yield (fp, readFileBinary $ templatedir </> fp)
+  runConduitRes $ src .| createTemplate .| sinkFile "rio.hsfiles"
```
roman: Oh 🙀🙀🙀 this is awesome! No longer nasty huge `hsfiles`, I'm copying this to my dotfiles like yesterday! No longer generating `Makefile`s with the wrong spacing 🎉

PR #88, opened 2018-05-05
Use ansi-terminal package, over raw ANSI codes

mpilgrem:
The advantage of the `ansi-terminal` package's `setSGRCode` function over the use of 'raw' ANSI codes is that, on Windows 10, the former will enable ANSI-capable ConHost terminal handles and, on legacy Windows, the former will emulate ANSI capabability on ConHost terminals (EDIT: actually, for the pure codes, the 'emulation' is to replace them with empty strings).

At the moment, `stack` v 1.7.1 outputs raw ANSI codes on Windows 10 because it is using RIO and neither `stack` nor RIO are enabling ANSI on ANSI-capable ConHost terminals. See https://github.com/commercialhaskell/stack/issues/3992.

### DIFF
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -13,6 +13,7 @@ extra-source-files:
 
 dependencies:
 - base >= 4.9 && < 10
+- ansi-terminal
 - bytestring
 - containers
 - deepseq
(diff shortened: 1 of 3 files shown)


mpilgrem:
Understood. Since I posted, I realised that the solution for `stack` (and other users of RIO) probably should sit 'above' RIO.

PR #90, opened 2018-05-21
Add support for monad-logger typeclassess for RIO

roman:
With this PR, we are going to be able to support libraries like persistent, which require `monad-logger` typeclass instances

### DIFF
--- a/rio-orphans/src/RIO/Orphans.hs
+++ b/rio-orphans/src/RIO/Orphans.hs
@@ -19,6 +19,10 @@ import Control.Monad.Trans.Resource.Internal (MonadResource (..), ReleaseMap, Re
 import Control.Monad.Trans.Resource (runResourceT)
 import Control.Monad.Trans.Control (MonadBaseControl (..))
 
+import qualified Control.Monad.Logger as LegacyLogger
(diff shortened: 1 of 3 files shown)


rio-orphans/src/RIO/Orphans.hs
```diff
@@ -35,6 +39,23 @@ instance MonadBaseControl IO (RIO env) where
   liftBaseWith = withRunInIO
   restoreM = return
 
+-- | @since 0.1.3.0
+instance Display LogStr where
+  display = displayBytesUtf8 . fromLogStr
+
+-- | @since 0.1.3.0
```
snoyberg: Can you add a package.yaml version bump and a note to the changelog?

rio-orphans/src/RIO/Orphans.hs
```diff
@@ -35,6 +39,23 @@ instance MonadBaseControl IO (RIO env) where
   liftBaseWith = withRunInIO
   restoreM = return
 
+-- | @since 0.1.3.0
```
snoyberg: Looks like these version numbers should be 0.1.1.0 instead.

roman: Good catch 👀

PR #91, opened 2018-05-21
Improvements in logging capabilities

roman:
* Add a `newLogFunc` function to allow the creation of LogFunc without the `with` semantics, a warning around async exceptions should be stated in the documentation, and recommendations should be to use `withLogFunc` unless the user knows what she is doing

* Add `newReloadableLogFunc`, which creates a `LogFunc` _and_ returns a `LogOptions -> IO ()` function that can be called to reload the settings of an existing logging.

I understand some of this changes are sensitive, and I would love to get a conversation going around those.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -351,23 +351,29 @@ getCanUseUnicode = do
 -- @since 0.0.0.0
 withLogFunc :: MonadUnliftIO m => LogOptions -> (LogFunc -> m a) -> m a
 withLogFunc options inner = withRunInIO $ \run -> do
-  if logTerminal options
-    then bracket
-            (newMVar mempty)
-            (\var -> do
-                state <- takeMVar var
-                unless (B.null state) (logSend options "\n"))
-            (\var -> run $ inner $ LogFunc
-                { unLogFunc = stickyImpl var options (simpleLogFunc options)
-                , lfOptions = Just options
-                }
-            )
-    else
-      run $ inner $ LogFunc
-        { unLogFunc = \cs src level str ->
-             simpleLogFunc options cs src (noSticky level) str
-        , lfOptions = Just options
-        }
+  bracket (newLogFunc options)
+          snd
+          (run . inner . fst)
+
+newLogFunc :: MonadIO m => LogOptions -> m (LogFunc, m ())
+newLogFunc options =
+  if logTerminal options then do
+    var <- newMVar mempty
+    return (LogFunc
+             { unLogFunc = stickyImpl var options (simpleLogFunc options)
+             , lfOptions = Just options
+             }
+           , do state <- takeMVar var
+                unless (B.null state) (liftIO $ logSend options "\n")
+           )
+  else
+    return (LogFunc
+            { unLogFunc = \cs src level str ->
+                simpleLogFunc options cs src (noSticky level) str
+            , lfOptions = Just options
+            }
+           , return ()
+           )
 
 -- | Replace Unicode characters with non-Unicode equivalents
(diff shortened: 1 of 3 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -8,7 +8,10 @@ module RIO.Prelude.Logger
   , logError
   , logOther
     -- * Running with logging
+  , newLogFunc
```
snoyberg: It doesn't seem strictly relevant to add in a `new` variant of these functions in addition to the `with`, unless I'm missing something.

rio/src/RIO/Prelude/Logger.hs
```diff
@@ -96,6 +106,13 @@ data LogFunc = LogFunc
   , lfOptions :: !(Maybe LogOptions)
   }
 
+-- | Pending documentation
+data ReloadableLogFunc = ReloadableLogFunc
+  { rlfLogFunc        :: !LogFunc
+  , rlfReloadLogFunc  :: !(LogOptions -> IO ())
```
snoyberg: I would have thought you'd instead have something like:

```haskell
newtype ReloadableLogFunc = ReloadableLogFunc (IORef (LogFunc, IO ()))

updateLogFunc :: ReloadableLogFunc -> LogFunc -> IO () -> IO ()
updateLogFunc (ReloadableLogFunc ref) lf cleanup = mask_ $ do
  (_, cleanupOld) <- readIORef ref
  cleanupOld
  writeIORef ref (lf, cleanup)

rlfLogFunc :: ReloadableLogFunc -> LogFunc
rlfLogFunc (ReloadableLogFunc ref) a b c d = do
  (lf, _) <- readIORef ref
  lf a b c d
```

rio/src/RIO/Prelude/Logger.hs
roman: I guess it depends on the use case; I'm using [other means ](https://github.com/roman/Haskell-componentm) to manage the lifecycle of the resources allocated by my application. I think this won't be much of a need if I can reload the `LogFunc` behavior after it is built.

rio/src/RIO/Prelude/Logger.hs
roman: I'm not opposed to this idea, but I believe there may be some changes needed to make this useful. The reason I used `LogOptions` as my input to reload, was because I can have that record outside of a `withLogFunc` block.

How would this update logic work if I only have a `withLogFunc`?

rio/src/RIO/Prelude/Logger.hs
snoyberg: I'd rather consider the addition of these functions separately then, and ideally figure out a way to avoid exposing them.

rio/src/RIO/Prelude/Logger.hs
snoyberg: Fairly poorly actually, but I think the whole approach has some problems around internal state. Let me try to reset the question a bit: is the log level the only thing we really need to be able to change? If so, I'd rather put just that in an `IORef` and make that directly modifiable via a setting to `LogOptions`.

rio/src/RIO/Prelude/Logger.hs
roman: I'll move this out of this PR and create another one with those changes only

rio/src/RIO/Prelude/Logger.hs
roman: Ah good point, as far as I'm aware if I can change `logLevel` and `verbosity` I think this should be good.

roman:
Closing this PR in favor of #93

PR #93, opened 2018-05-23
Allow LogFunc to reload LogLevel and verbose options dynamically

roman:
These changes are an improvement over PR #91

### DIFF
--- a/rio/test/RIO/LoggerSpec.hs
+++ b/rio/test/RIO/LoggerSpec.hs
@@ -24,3 +24,13 @@ spec = do
       logStickyDone "XYZ"
     builder <- readIORef ref
     toLazyByteString builder `shouldBe` "ABC\b\b\b   \b\b\bshould appear\nABC\b\b\b   \b\b\bXYZ\n"
+  it "updateLogMinLevel" $ do
+    (ref, options) <- logOptionsMemory
+    withLogFunc (options & setLogMinLevel LevelDebug)
+      $ \lf -> runRIO lf $ do
+        logDebug "should appear"
+        -- reset log min level to info
+        updateLogMinLevel lf LevelInfo
+        logDebug "should not appear"
+    builder <- readIORef ref
+    toLazyByteString builder `shouldBe` "should appear\n"
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 module RIO.Prelude.Logger
@@ -20,6 +21,8 @@ module RIO.Prelude.Logger
   , setLogUseTime
   , setLogUseColor
   , setLogUseLoc
+  , updateLogMinLevel
+  , updateLogVerboseFormat
     -- * Advanced logging functions
     -- ** Sticky logging
   , logSticky
@@ -94,16 +97,20 @@ instance HasLogFunc LogFunc where
 data LogFunc = LogFunc
   { unLogFunc :: !(CallStack -> LogSource -> LogLevel -> Utf8Builder -> IO ())
   , lfOptions :: !(Maybe LogOptions)
+  , lfUpdateLogOptions :: !((LogOptions -> LogOptions) -> IO ())
   }
 
 -- | Perform both sets of actions per log entry.
 --
 -- @since 0.0.0.0
 instance Semigroup LogFunc where
-  LogFunc f o1 <> LogFunc g o2 = LogFunc
+
+  LogFunc f o1 update1 <> LogFunc g o2 update2 = LogFunc
     { unLogFunc = \a b c d -> f a b c d *> g a b c d
     , lfOptions = o1 `mplus` o2
+    , lfUpdateLogOptions = \optionMod -> update1 optionMod *> update2 optionMod
     }
+
 
 -- | 'mempty' peforms no logging.
 --
@@ -116,7 +123,7 @@ instance Monoid LogFunc where
 --
 -- @since 0.0.0.0
 mkLogFunc :: (CallStack -> LogSource -> LogLevel -> Utf8Builder -> IO ()) -> LogFunc
-mkLogFunc f = LogFunc f Nothing
+mkLogFunc f = LogFunc f Nothing (\_ -> return ())
 
 -- | Generic, basic function for creating other logging functions.
 --
@@ -128,8 +135,8 @@ logGeneric
   -> Utf8Builder
   -> m ()
 logGeneric src level str = do
-  LogFunc logFunc _ <- view logFuncL
-  liftIO $ logFunc callStack src level str
+  LogFunc {unLogFunc} <- view logFuncL
+  liftIO $ unLogFunc callStack src level str
 
 -- | Log a debug level message with no source.
 --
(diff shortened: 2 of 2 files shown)


rio/test/RIO/LoggerSpec.hs
```diff
@@ -24,3 +24,13 @@ spec = do
       logStickyDone "XYZ"
     builder <- readIORef ref
     toLazyByteString builder `shouldBe` "ABC\b\b\b   \b\b\bshould appear\nABC\b\b\b   \b\b\bXYZ\n"
+  it "updateLogMinLevel" $ do
+    (ref, options) <- logOptionsMemory
+    withLogFunc (options & setLogMinLevel LevelDebug)
+      $ \lf -> runRIO lf $ do
+        logDebug "should appear"
+        -- reset log min level to info
+        updateLogMinLevel lf LevelInfo
```
roman: Note: This test is not as realistic, the `updateLogMinLevel` function will be likely called from a different environment than the logging environment.

snoyberg:
What about something with less surface area of change, e.g.

```diff
diff --git a/rio/src/RIO/Prelude/Logger.hs b/rio/src/RIO/Prelude/Logger.hs
index 0e32e6e..f616493 100644
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -15,6 +15,7 @@ module RIO.Prelude.Logger
     -- ** Log options
   , LogOptions
   , setLogMinLevel
+  , setLogMinLevelIO
   , setLogVerboseFormat
   , setLogTerminal
   , setLogUseTime
@@ -274,7 +275,7 @@ logOptionsMemory :: MonadIO m => m (IORef Builder, LogOptions)
 logOptionsMemory = do
   ref <- newIORef mempty
   let options = LogOptions
-        { logMinLevel = LevelInfo
+        { logMinLevel = return LevelInfo
         , logVerboseFormat = False
         , logTerminal = True
         , logUseTime = False
@@ -299,7 +300,7 @@ logOptionsHandle handle' verbose = liftIO $ do
   useUtf8 <- canUseUtf8 handle'
   unicode <- if useUtf8 then return True else getCanUseUnicode
   return LogOptions
-    { logMinLevel = if verbose then LevelDebug else LevelInfo
+    { logMinLevel = return $ if verbose then LevelDebug else LevelInfo
     , logVerboseFormat = verbose
     , logTerminal = terminal
     , logUseTime = verbose
@@ -385,7 +386,7 @@ noSticky level = level
 --
 -- @since 0.0.0.0
 data LogOptions = LogOptions
-  { logMinLevel :: !LogLevel
+  { logMinLevel :: !(IO LogLevel)
   , logVerboseFormat :: !Bool
   , logTerminal :: !Bool
   , logUseTime :: !Bool
@@ -401,7 +402,10 @@ data LogOptions = LogOptions
 --
 -- @since 0.0.0.0
 setLogMinLevel :: LogLevel -> LogOptions -> LogOptions
-setLogMinLevel level options = options { logMinLevel = level }
+setLogMinLevel level options = options { logMinLevel = return level }
+
+setLogMinLevelIO :: IO LogLevel -> LogOptions -> LogOptions
+setLogMinLevelIO level options = options { logMinLevel = level }
 
 -- | Use the verbose format for printing log messages.
 --
@@ -446,8 +450,9 @@ setLogUseLoc :: Bool -> LogOptions -> LogOptions
 setLogUseLoc l options = options { logUseLoc = l }
 
 simpleLogFunc :: LogOptions -> CallStack -> LogSource -> LogLevel -> Utf8Builder -> IO ()
-simpleLogFunc lo cs _src level msg =
-    when (level >= logMinLevel lo) $ do
+simpleLogFunc lo cs _src level msg = do
+    minLevel <- logMinLevel lo
+    when (level >= minLevel) $ do
       timestamp <- getTimestamp
       logSend lo $ getUtf8Builder $
         timestamp <>
@@ -546,13 +551,15 @@ stickyImpl ref lo logFunc loc src level msgOrig = modifyMVar_ ref $ \sticky -> d
       let bs = toStrictBytes $ toLazyByteString $ getUtf8Builder msgOrig
       logSend lo (byteString bs <> flush)
       return bs
-    _
-      | level >= logMinLevel lo -> do
+    _ -> do
+      minLevel <- logMinLevel lo
+      if level >= minLevel
+        then do
           clear
           logFunc loc src level msgOrig
           unless (B.null sticky) $ logSend lo (byteString sticky <> flush)
           return sticky
-      | otherwise -> return sticky
+        else return sticky
 
 -- | Is the log func configured to use color output?
 --
```

roman:
You are right; this approach has a better footprint regarding LOC changes. To see if I understand how I would use it, I'll write a quick example:

__DISCLAIMER__: Code not tested in REPL

```haskell
data App = { appLogFunc :: !LogFunc, appUpdateLogLevel :: !(LogLevel -> IO ()) }
logLevelRef <- newIORef LevelDebug
let updateLogLevel level = atomicModifyIORef' logLevelRef (\_ -> (level, ()))
logOptions <- setLogMinLevelIO (readIORef logLevelRef) <$> logOptionsHandle stdout False
withLogFunc logOptions $ \logFunc ->
  runMain (App logFunc updateLogLevel)
```

With the `setLogMinLevelIO` solution: 

* I need to maintain some mutable variable (e.g. `logLevelRef`) besides the `LogFunc` in the environment of my application to be sure I can update it. 

* We'll need to add more `IORef` entries (and as a user of the lib, maintain those `IORef`s) to dynamically update other settings in the logging system (e.g. `verbose`)

Between maintaining more variables (e.g. `logLevelRef`) on my app state and adding more footprint to the `RIO.Prelude.Logger`, I think I prefer the latter.

The fact that you are a bit hesitant about the changes in the PR makes me believe I have a blind spot about tradeoffs somewhere with the implementation I shared, would you like to share gut feelings about things to be sure?

roman:
>  For the PR you've sent now, what happens if you try to change between terminal and non-terminal logging?

I completely agree that is a *bad* idea to change terminal to non-terminal logging (the previous PR had that issue and I added a comment for that); the good news is,  that won't be something that would be possible to do here as long as we:

* Keep the `LogFunc` constructor private
* Don't export the `updateLogOptions` function

The only way to call that `lfUpdateLogOptions`  attribute from `LogFunc`  is if we provide a public interface for that, as we explicitly do with `updateLogMinLevel` and `updateLogVerboseFormat`, those are the only two attributes that may be updated.

> More generally, I'm unsure if the changes you're providing work, whereas the change I showed above is much easier to analyze and be sure it's not affecting the rest of the program.

I understand the tests I added are probably not enough to prove the absence of bugs or performance degradation, I can make an attempt at extending those a bit more if you want me to.

> If you really want to, you can create a helper datatype which captures all of the things that can be updated at runtime.

So that would extend the footprint at the library level, or is this something the client of the library will have to manage?

---

Adding points to your implementation, I appreciate that the change here avoids accidental future changes in the API that will expose a way to be able to update all options. I was expecting this to be captured on code reviews though.

Apologies for being a bit pushy here, I really want this API to be top-notch and I appreciate your time in making so.

snoyberg:
I'm always highly nervous of any approach that requires that bugs will "be captured on code reviews." In other words, I consider this approach too fragile. Marking the fields that can safely mutate over time as mutable is a straightforward thing to justify.

To be honest, I'm not convinced at all of the concern of putting too much pressure on the application implementer with the approach I shared. There's already plenty of code necessary to parse out and reload settings. Having to create 3 `IORef`s doesn't seem too bad to me. I have a strong suspicion that if you actually compare the application code for the various API options, they'll come out to just about the same SLOC.

roman:
> To be honest, I'm not convinced at all of the concern of putting too much pressure on the application implementer with the approach I shared

Fair, I'm happy just to reload `minLogLevel` and `verboseFormat` for now, I'll go ahead and create a new PR with your suggested changes. I'll work on it tonight (PST)

roman:
Closing this PR in favor of #94

PR #94, opened 2018-05-25
Allow reload for logMinLevel and logVerboseFormat settings

roman:
Opening this PR in favor of #93

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -15,7 +15,9 @@ module RIO.Prelude.Logger
     -- ** Log options
   , LogOptions
   , setLogMinLevel
+  , setLogMinLevelIO
   , setLogVerboseFormat
+  , setLogVerboseFormatIO
   , setLogTerminal
   , setLogUseTime
   , setLogUseColor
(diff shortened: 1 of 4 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -401,15 +403,35 @@ data LogOptions = LogOptions
 --
 -- @since 0.0.0.0
 setLogMinLevel :: LogLevel -> LogOptions -> LogOptions
-setLogMinLevel level options = options { logMinLevel = level }
+setLogMinLevel level options = options { logMinLevel = return level }
+
+-- | Set the minimum log level. Messages below this level will not be
+-- printed.This modifier allows to alter the verbose format value dynamically at
```
snoyberg: Needs a space between `printed.` and `This`. Also, alternative text could simply be to refer to `setLogLevelIO` and explain that it allows dynamic updates at runtime.

rio/src/RIO/Prelude/Logger.hs
```diff
@@ -401,15 +403,35 @@ data LogOptions = LogOptions
...
+-- runtime.
+--
+-- Default: in verbose mode, 'LevelDebug'. Otherwise, 'LevelInfo'.
+--
+-- @since 0.1.2.0
```
snoyberg: Please update the ChangeLog and package.yaml file too.

PR #95, opened 2018-05-26
Allow alloc of LogFunc outside a with block

roman:
This pull request adds the function `newLogFunc` to the `RIO.Prelude.Logger` module, this function returns a tuple with a `LogFunc` and a teardown sub-routine. This function is useful when using a mechanism other than `bracket` (e.g. [`componentm`](https://github.com/roman/Haskell-componentm/tree/master/componentm)) to deal with resource management.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -8,8 +8,8 @@ module RIO.Prelude.Logger
   , logError
   , logOther
     -- * Running with logging
-  , newLogFunc
   , withLogFunc
+  , newLogFunc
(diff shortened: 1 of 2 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -8,6 +8,7 @@ module RIO.Prelude.Logger
   , logError
   , logOther
     -- * Running with logging
+  , newLogFunc
```
snoyberg: Minor point, but can you add this export _below_ `withLogFunc`, since that's what we want people using in general?

Issue #82, opened 2018-04-09
Offer safe variants for the RIO.List.Partial functions in the RIO.List module


roman:
Iterating over the work from #43, what do you think about adding all the functions from `RIO.List.Partial` with a `safe` prefix in the `RIO.List` module?

There is no clear replacement for the functions in `RIO.List.Partial`, and I would avoid them as much as possible if I can.

snoyberg:
In principle: I'm OK with this approach. I'm not convinced that `safe` will be the right prefix in all cases. For example: `maximum` would probably be better replaced by `maximumMay` or `maximumMaybe` to match existing naming conventions IMO.

mitchellwrosen:
FYI: `scanl1`, `scanr1` aren't partial

snoyberg:
Implemented in #96

tejon:
Late entry, but perhaps relevant... yesterday I had to switch some lists over to vectors so I could use `statistics` operations on them. I had been using some of the (list-specific) `*Def` functions from `safe`, and needed to replicate those; and some of the `statistics` operations I wanted were partial on empty vectors, too, which needed fixing. I wound up writing this:

```haskell
defaulting :: Foldable f => b -> (f a -> b) -> f a -> b
defaulting d f xs
  | null xs   = d
  | otherwise = f xs
```

So now I have e.g. `defaulting 0 Vector.maximum`. The `*Maybe` pattern is a trivial special case:

```haskell
mayhap :: Foldable f => (f a -> b) -> f a -> Maybe b
mayhap f = defaulting Nothing (Just . f)
```

Are there any particular downsides to this technique? It seems far more elegant than a slew of specialized, type-specific functions.

PR #96, opened 2018-05-26
Alternative (safe) functions for RIO.List.Partial

roman:
This PR adds safe versions of partial functions on lists. This will close ticket #82

### DIFF
--- a/rio/src/RIO/List/Partial.hs
+++ b/rio/src/RIO/List/Partial.hs
@@ -20,6 +20,9 @@ module RIO.List.Partial
   -- * Building lists
 
   -- ** Scans
+  --
+  -- These functions are not partial, they are being exported here for legacy
+  -- reasons, they may be removed from this module on a future major release
   , Data.List.scanl1
   , Data.List.scanr1
 
@@ -28,4 +31,3 @@ module RIO.List.Partial
   ) where
 
 import qualified Data.List
-
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -4,6 +4,9 @@
 
 * Add `newLogFunc` function to create `LogFunc` records outside of a callback scope
 * Allow dynamic reloading of `logMinLevel` and `logVerboseFormat` for the `LogOptions` record
+* Add `headMaybe`, `lastMaybe`, `tailMaybe`, `initMaybe`, `maximumMaybe`, `minimumMaybe`,
+  `maximumByMaybe`, `minimumByMaybe` functions to `RIO.List` module (issue #82)
+* Move non partial functions `scanr1` and `scanl1` from `RIO.List.Partial` to `RIO.List` (issue #82)
 
 ## 0.1.2.0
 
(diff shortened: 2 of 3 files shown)


tejon:
I'm not especially opinionated on this, but someone needs to whip out the paint cans: `safe`, an exceedingly popular module (300+ revdeps including `stack`), chose to use `*May` rather than `*Maybe` for these functions. If we choose not to follow that lead, we should at least be deliberate and explicit about it.

roman:
Agreed, don't assume I'm pushing for any particular name scheme, I just did a proposal, and I'm very open to change it to whatever makes sense

rio/src/RIO/List/Partial.hs
```diff
@@ -17,15 +17,8 @@ module RIO.List.Partial
   , Data.List.maximumBy
   , Data.List.minimumBy
 
-  -- * Building lists
-
-  -- ** Scans
-  , Data.List.scanl1
```
snoyberg: For backwards compat, these should remain here, but a comment explaining that they should be removed in the future would be a good thing.

tejon:
I wasn't sure if `readMaybe` was a solitary witness; it's not, there's also `bitSizeMaybe` (perhaps others, I wasn't thorough). I think that's a strong enough pattern to follow.

Another shade from `base` is `try*`, e.g. `tryReadMVar`. But despite tying for shortest, I like it less for a couple of reasons. First, autocompletion sorting; second, `try` is associated with exceptions and I'd rather not muddy that.

So, I'm up to a solid +0.5 for `*Maybe`. If @roman is at least +0.3, we've got a whole +1!

rio/src/RIO/List/Partial.hs
```diff
@@ -17,15 +17,20 @@ module RIO.List.Partial
   , Data.List.maximumBy
   , Data.List.minimumBy
 
-  -- * Building lists
-
   -- ** Scans
-  , Data.List.scanl1
-  , Data.List.scanr1
+  , RIO.List.Partial.scanl1
```
snoyberg: This is actually still a breaking change, since now the identifier can conflict with one imported from elsewhere (like the one in `RIO.List`, meaning this will almost certainly break code).

roman: 🤦‍♂️

rio/src/RIO/List/Partial.hs
```diff
@@ -17,8 +17,6 @@ module RIO.List.Partial
   , Data.List.maximumBy
   , Data.List.minimumBy
 
-  -- * Building lists
```
snoyberg: There are still changes present in this file versus master, which probably shouldn't be here. Also, I _do_ think it's worth adding a comment that `scanl1` and `scanr1` are here for only historical reasons.

roman:
@snoyberg I did a rebase on the PR to edit the original commit, also removed the reverted changes, the branch should be cleaner now

PR #98, opened 2018-05-27
Add runRIOSimple/SimpleApp

snoyberg:


rio/src/RIO/Prelude/Simple.hs
```diff
@@ -0,0 +1,54 @@
...
+--
+-- * Logging to stderr
+--
+-- * If the @RIO_VERBOSE@ environment variable is set, turns on
+--   verbose logging
+--
+-- * Default process context
+--
+-- @since 0.1.3.0
+runRIOSimple :: MonadIO m => RIO SimpleApp a -> m a
```
roman: I know is silly, but I dislike the `OS` in `runRIOSimple`, would it be too much just to calling it `runSimpleApp` instead?

Also, why is it logging by default to `stderr` and not to `stdout`?

PR #101, opened 2018-05-29
Various logger extensions

roman:
Descriptions of rationale are on commit messages.

This is the second iteration over pull request #100

### DIFF
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -5,6 +5,7 @@
 * Add `newLogFunc` function to create `LogFunc` records outside of a callback scope
 * Add `logGenericCallStack` to be able to register a log entry `CallStack` from a
   different thread
+* Add `getLogMinLevel` to gather what is the log min level from `LogOptions`
 * Allow dynamic reloading of `logMinLevel` and `logVerboseFormat` for the `LogOptions` record
 
 ## 0.1.2.0
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -3,6 +3,8 @@
 ## 0.1.3.0
 
 * Add `newLogFunc` function to create `LogFunc` records outside of a callback scope
+* Add `logGenericCallStack` to be able to register a log entry `CallStack` from a
+  different thread
 * Allow dynamic reloading of `logMinLevel` and `logVerboseFormat` for the `LogOptions` record
 
 ## 0.1.2.0
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -15,7 +15,7 @@ module RIO.Prelude.Logger
   , logOptionsHandle
     -- ** Log options
   , LogOptions
-  , getLogMinLevel
+  , checkShouldLog
   , setLogMinLevel
   , setLogMinLevelIO
   , setLogVerboseFormat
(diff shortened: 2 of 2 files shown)


snoyberg:
The issue with both of these is that they constrain the extensibility of the library going forward. What if, e.g., we want to provide the ability for the user to give us a `LogLevel -> Bool` predicate instead of a simple minimum? That's not crazy; with custom log levels, it might be warranted.

If the goal is still the thread ID issue, here's what I'd had in mind:

```haskell
setModifyLogMessage
  :: (Utf8Builder -> IO Utf8Builder) -- ^ guaranteed to be run in original thread
  -> LogOptions
  -> LogOptions
```

roman:
I like the idea of `setModifyLogMessage`; however, I think what I'm trying to solve in this two commits is a different challenge.

In regards to the first commit:

I'm building a `LogFunc` via [`mkLogFunc`](https://github.com/roman/Haskell-capataz/blob/improve-logger-performance/src/Control/Concurrent/Capataz/Util/Logger.hs#L77), that decorates a regular RIO `LogFunc`. This regular `LogFunc` is running on a separate dedicated thread for logging. 

Imagine my custom `LogFunc` gets all the arguments and send them through a queue to a different thread that is using the original RIO's `LogFunc` and calls `logGeneric` call with the arguments. Currently, there is no way for me to change the `CallStack` value (`logGeneric` gets it call stack from a `HasCallStack` constraint), so all the logging we do on the dedicated thread is going to show me the _wrong_ stack, making the feature useless.
In the second commit:

Given I'm sending the log messages over a queue, it is wasteful to send messages to discover at the receiving end that the log message is not going to be used at all, this situation is particularly dangerous in an app with more than a hundred of threads logging values. The `getLogMinLevel` function allows me to [do the check](https://github.com/roman/Haskell-capataz/blob/improve-logger-performance/src/Control/Concurrent/Capataz/Util/Logger.hs#L78) that we are doing in the `simpleLogFunc` implementation to avoid that extra hop.

I'm 100% in for keeping the resolving of the log options internal. I would like a query function to be available in `LogOptions` that keep those interior details opaque (either by having this function returning a Bool, or receiving a block that gets executed when conditions are met). 

This change makes `getLogMinLevel`  an alias to `logMinLevel` accessor of `LogOptions`, but, that does not need to be the case, we are open to change the resolution of the `logMinLevel` because `getLogMinLevel` will meet the same interface by different means.

I hope I'm able to explain my rational effectively and I'm not missing anything obvious here.

roman:
After giving it more thought, and, if I understand you correctly, the `getLogMinLevel` exposure is not great because we want to be able to use other arguments of the `LogFunc` signature to make a decision about what is the log min level. I think this is fair, my goal though is to create "middleware" to `LogFunc` values, so that we can make decisions in any step of the chain.

Assuming we require _all_ the values to resolve the `LogMinLevel`, would this be ok?

```haskell
type LogFuncF = CallStack -> LogSource -> LogLevel -> Utf8Builder

isValidLogLevel :: LogOptions -> LogFuncF -> IO Bool
isValidLogLevel lo _cs _source currentLevel _payload = do
  minLevel <- logMinLevel lo
  -- Later extend with future logic for deciding min level
  return (currentLevel >= minLevel) 
```

We then would use this function [here](https://github.com/commercialhaskell/rio/blob/master/rio/src/RIO/Prelude/Logger.hs#L487) and [here](https://github.com/commercialhaskell/rio/blob/master/rio/src/RIO/Prelude/Logger.hs#L592)

roman:
Hey @snoyberg, I implemented a version of your suggestion, the signature is different, it does not work inside the monadic context, but rather, is a pure function that builds a new `LogFunc` from an old one.

The reason I changed the signature, is because I'm building a new `LogFunc` decorator _outside_ of my business layer, so having the `m` context there doesn't help.

The [documentation of the function](https://github.com/commercialhaskell/rio/pull/101/commits/a0057193fd5e62cff94939279d658f598d929a61#diff-98e11794ea9b8ffd9e39af5bf3ef3bd7R439) has an example of how you would use it.

Personally, I'm not thrilled with this implementation, I have the impression that this should be simpler.

Maybe a possible solution could be to provide primitives that decorate `LogFunc` values in different ways and use those internally as well, an example serves best:

```haskell
filterMinLogLevel :: LogLevel -> LogFunc -> LogFunc
filterMinLogLevel minLevel (LogFunc {unLogFunc}) = mkLogFunc $ \cs src level msg -> do
  when (level >= minLevel) (unLogFunc cs src level msg)

appendTimestamp :: LogFunc -> LogFunc
appendTimestamp (LogFunc {unLogFunc}) = mkLogFunc $ \cs src level msg -> do
    timestamp <- getTimestamp
    unlogFunc cs src level (timestamp <> msg)
  where
    getTimestamp = ...
  
simpleLogFunc :: LogOptions -> LogFunc -> LogFunc
simpleLogFunc lo baseLogFunc = mkLogFunc $ \cs src level msg ->
    minLevel <- logMinLevel lo
    isVerbose <- logVerboseFormat lo
    foldl step baseLogFunc [(filterMinLogLevel minLevel, True), (appendTimestamp, isVerbose), ...]
  where
    step accLogFunc (modLogFunc, shouldApply)
       | shouldApply = modLogFunc accLogFunc
       | otherwise = accLogFunc
```

I like the approach above because we define all the properties of `simpleLogFunc` as combinators that we can merge together, and we offer the possibility for others to build their own version `LogOptions` (if they need to) and keep using our simple primitives.

What do you think?

roman:
Closing this ticket given it adds unnecessary complexity to the library.

PR #105, opened 2018-06-06
Add instances for MonadWriter and MonadState with some reference 

mauropalsgraaf:


rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +48,60 @@ instance MonadUnliftIO (RIO env) where
...
+ioRefToSomeRef :: IORef a -> SomeRef a
+ioRefToSomeRef ref = do
+  SomeRef (readIORef ref)
+          (\val -> atomicModifyIORef' ref (\old -> (val, ())))
+
+uRefToSomeRef :: Unbox a => URef RealWorld a -> SomeRef a
+uRefToSomeRef ref = do
+  SomeRef (readURef ref) (writeURef ref)
+
+instance MonadState s (RIO (SomeRef s)) where
```
snoyberg:
```haskell
instance HasStateRef s env => MonadState s (RIO env) where
  get = do
    ref <- view stateRefL
    liftIO $ readSomeRef ref
```

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +48,60 @@ instance MonadUnliftIO (RIO env) where
...
+writeSomeRef (SomeRef _ x) = x
+
+modifySomeRef :: SomeRef a -> (a -> a) -> IO ()
+modifySomeRef (SomeRef read write) f =
+  (f <$> read) >>= write
+
+ioRefToSomeRef :: IORef a -> SomeRef a
+ioRefToSomeRef ref = do
+  SomeRef (readIORef ref)
+          (\val -> atomicModifyIORef' ref (\old -> (val, ())))
```
snoyberg: I don't believe the `atomic` adds anything here, since you're ignoring the old value anyway.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +48,60 @@ instance MonadUnliftIO (RIO env) where
 instance PrimMonad (RIO env) where
     type PrimState (RIO env) = PrimState IO
     primitive = RIO . ReaderT . const . primitive
+
+data SomeRef a
+  = SomeRef !(IO a) !(a -> IO ())
+
+readSomeRef :: SomeRef a -> IO a
```
snoyberg: We should have haddocks on all of the newly exposed identifiers.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -51,67 +50,47 @@ instance PrimMonad (RIO env) where
...
 
-
 instance MonadState s (RIO (SomeRef s)) where
   get = ask >>= liftIO . readSomeRef
   put st = do
     ref <- ask
     liftIO $ writeSomeRef ref st
 
-instance Monoid w => MonadWriter w (RIO (SomeRef w)) where
+instance (Show w, Monoid w) => MonadWriter w (RIO (SomeRef w)) where
```
roman: Why do we need the `Show` constraint here?

mauropalsgraaf: That was for debugging and i forgot to remove 👍

mauropalsgraaf:
So I've updated the comments. I've also added the HasWriteRef typeclass which has the same purpose as the HasLogFunc and HasStateRef. Regarding haddock, i was not sure what to put as documentation for the instances MonadState and MonadWriter. Lastly, I assumed the next version is going to be 0.1.3.0, if anything else was in mind i will change the version number.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+          (\val -> modifyIORef' ref (\_ -> val))
+
+uRefToSomeRef :: Unbox a => URef RealWorld a -> SomeRef a
+uRefToSomeRef ref = do
+  SomeRef (readURef ref) (writeURef ref)
+
+-- | Environment values with stateful capabilities to SomeRef
+--
+-- @since 0.1.3.0
+class HasStateRef s env | env -> s where
```
snoyberg: I don't think this class has been exported.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+-- | Identity state reference where the SomeRef is the env
+--
+-- @since 0.1.3.0
+instance HasStateRef a (SomeRef a) where
+  stateRefL = lens id (\_ x -> x)
+
+-- | Environment values with writing capabilities to SomeRef
+--
+-- @since 0.1.3.0
+class HasWriteRef w env | env -> w where
```
snoyberg: Same with this class.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+  writeRefL :: Lens' env (SomeRef w)
+
+-- | Identity write reference where the SomeRef is the env
+--
+-- @since 0.1.3.0
+instance HasWriteRef a (SomeRef a) where
```
roman: @mauropalsgraaf  wondering what is the reason we have a different typeclass for the same purpose?

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
 instance PrimMonad (RIO env) where
     type PrimState (RIO env) = PrimState IO
     primitive = RIO . ReaderT . const . primitive
+
+-- | Abstraction over how to read from and write to a reference
```
roman: perhaps?

s/a reference/a mutable reference/

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+--
+-- @since 0.1.3.0
+data SomeRef a
+  = SomeRef !(IO a) !(a -> IO ())
+
+-- | read from a SomeRef
```
roman: s/read/Read/

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+--
+-- @since 0.1.3.0
+readSomeRef :: SomeRef a -> IO a
+readSomeRef (SomeRef x _) = x
+
+-- | write to a SomeRef
```
roman: s/write/Write/

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +50,104 @@ instance MonadUnliftIO (RIO env) where
...
+--
+-- @since 0.1.3.0
+writeSomeRef :: SomeRef a -> a -> IO ()
+writeSomeRef (SomeRef _ x) = x
+
+-- | modify a SomeRef
```
roman: s/modify/Modify/

rio/src/RIO/Prelude/RIO.hs
mauropalsgraaf: If you would only use one typeclass, the somerefs have to always be the same, while now I can have an environment with two SomeRefs, one for writing and one for state. We could add a default instance something like `instance HasStateRef w env => HasWriteRef w env where` to have both on the same place right?

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -1,11 +1,28 @@
...
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
 module RIO.Prelude.RIO
   ( RIO (..)
   , runRIO
   , liftRIO
+  -- * SomeRef for Writer/State interfaces
+  , SomeRef
+  , HasStateRef
```
snoyberg: Both this and the following need to be exported with the member functions, e.g. `HasStateRef (..)`.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +52,104 @@ instance MonadUnliftIO (RIO env) where
...
+-- | Abstraction over how to read from and write to a mutable reference
+--
+-- @since 0.1.3.0
+data SomeRef a
+  = SomeRef !(IO a) !(a -> IO ())
+
+-- | Read from a SomeRef
+--
+-- @since 0.1.3.0
+readSomeRef :: SomeRef a -> IO a
```
snoyberg: Only noticing this now: for consistency with the rest of rio, this should be lifted to `MonadIO`.

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +52,104 @@ instance MonadUnliftIO (RIO env) where
...
+readSomeRef (SomeRef x _) = x
+
+-- | Write to a SomeRef
+--
+-- @since 0.1.3.0
+writeSomeRef :: SomeRef a -> a -> IO ()
```
snoyberg: Please lift to `MonadIO`

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +52,104 @@ instance MonadUnliftIO (RIO env) where
...
+writeSomeRef (SomeRef _ x) = x
+
+-- | Modify a SomeRef
+--
+-- @since 0.1.3.0
+modifySomeRef :: SomeRef a -> (a -> a) -> IO ()
```
snoyberg: Please lift to `MonadIO`

rio/src/RIO/Prelude/RIO.hs
```diff
@@ -35,3 +52,104 @@ instance MonadUnliftIO (RIO env) where
...
+readSomeRef :: SomeRef a -> IO a
+readSomeRef (SomeRef x _) = x
+
+-- | Write to a SomeRef
+--
+-- @since 0.1.3.0
+writeSomeRef :: SomeRef a -> a -> IO ()
+writeSomeRef (SomeRef _ x) = x
+
+-- | Modify a SomeRef
```
snoyberg: Can you add a comment that this is subject to race conditions due to lack of atomic operations?

rio/src/RIO/Prelude/Reexports.hs
```diff
@@ -50,6 +50,8 @@ module RIO.Prelude.Reexports
   , Control.Monad.Reader.MonadReader
   , Control.Monad.Reader.MonadTrans(..)
   , Control.Monad.Reader.ReaderT(..)
+  , Control.Monad.State.MonadState(..)
```
snoyberg: Since the method names here can conflict with other libraries in some cases, I'd rather not include them in the `RIO` module itself. We could add a `RIO.State`/`RIO.Writer` pair of modules, though.

roman: Wondering if it is worth adding a `Monad` in the name? I'm not sure, I guess we are almost certain there is never going to be something called `State` or `Writer` that doesn't imply the monads

rio/src/RIO/Prelude/RIO.hs
mauropalsgraaf: done

rio/src/RIO/Prelude/RIO.hs
mauropalsgraaf: Makes sense, I'm wondering though if the definition of SomeRef is still oke? Wouldn't it be better if the SomeRef knew something about MonadIO instead of being (SomeRef !(IO a) !(a -> IO a))? As far as i know, we cannot add a typeclass constraint on the creation of the datatype but i found this post that says its possible with GADTs (https://wiki.haskell.org/Data_declaration_with_constraint)

rio/src/RIO/Prelude/RIO.hs
mauropalsgraaf: done

rio/src/RIO/Prelude/RIO.hs
mauropalsgraaf: done

rio/src/RIO/Prelude/Reexports.hs
mauropalsgraaf: @snoyberg So if I understand correctly,  you would like to have a new module RIO.State & RIO.Writer (next to the RIO now which contains everything) that only contains the State and Writer monad stuff created in this PR? So that would mean to exclude the SomeRef, HasStateRef, MonadState from the main RIO module and only import these in two new modules (RIO.State and RIO.Writer)?

snoyberg: I meant only the `MonadState (..)` and `MonadWriter (..)` reexports, which will cause name collisions, which we want to avoid in a prelude-like module. There's no problem with everything else that's been exported.

mauropalsgraaf: I think I still don't get it, but I've added a commit to make it more visible. If i remove the MonadState deps from the Reexports, then I certainly need to include them in the RIO.Prelude.RIO file, otherwise MonadState / MonadWriter is not found. Now, I've also created the RIO.State and RIO.Writer which simply exposes the Control.Monad.State.MonadState and Control.Monad.Writer.MonadWriter. Does this makes sense? The changes can be seen in the commit.

snoyberg: Yes, I think that addresses it. It avoids an import of the `RIO` module from including the `get` identifier, for example, which conflicts with some other libraries.

snoyberg: Unfortunately, for some reason, that change is not appearing in the diff for this PR, and I can't understand why.

snoyberg:
I checked on this locally, and the latest commits appeared. Strange bug in the Github interface I guess. Thanks for the PR!

PR #109, opened 2018-06-19
Reexport Control.Monad.Reader.Reader, runReader

tejon:
Not strictly necessary if the reexport of Identity is adopted, but saves a bit of boilerplate.


snoyberg:
I'm OK with this, can you add a ChangeLog entry?

PR #110, opened 2018-06-19
Reexport class GHC.Records.HasField +methods

tejon:
Just a suggestion. This seems like the cleanest fully-standard way of disambiguating `DuplicateRecordFields`. Does have a few name conflicts on Stackage. Not sure what the floor on RIO's GHC version support is, if it's pre-8.0 this is a non-starter.

### DIFF
--- a/rio/src/RIO/Prelude/Reexports.hs
+++ b/rio/src/RIO/Prelude/Reexports.hs
@@ -179,6 +179,7 @@ module RIO.Prelude.Reexports
   , Data.Word.byteSwap64
   , Foreign.Storable.Storable
   , GHC.Generics.Generic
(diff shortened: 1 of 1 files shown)


snoyberg:
We only support GHC 8 and up, but it looks like this fails anyway:

https://travis-ci.org/commercialhaskell/rio/jobs/394275383

Nonetheless, I'm weakly -1 on this, if only because I've never used this class personally, and therefore am not sure if it really needs to be that readily accessible.

tejon:
Huh, I guess this didn't actually drop until 8.2. Thought I saw documentation saying otherwise, but clearly was mistaken!

PR #112, opened 2018-06-24
Reexport `Data.Text.replace` and `Data.Text.Lazy.replace`

markus1189:
I needed that function today and saw that it is not (yet) exposed.  So I went ahead and did it :)


tejon:
These functions are exported from `.Partial`.

markus1189:
Indeed! Sorry my bad

PR #115, opened 2018-07-09
Export Prelude.fromEnum from RIO, export fromJust, read and toEnum form RIO.Partial

locallycompact:
Hi, I noticed these are missing when I tried to use them.

### DIFF
--- a/rio/src/RIO/Partial.hs
+++ b/rio/src/RIO/Partial.hs
@@ -1,0 +1,10 @@
+-- | Partial functions.
+--
+module RIO.Partial
+  ( Data.Maybe.fromJust
+  , Prelude.read
+  , Prelude.toEnum
+  ) where
+
+import qualified Data.Maybe
+import qualified Prelude
--- a/rio/src/RIO/Prelude/Reexports.hs
+++ b/rio/src/RIO/Prelude/Reexports.hs
@@ -211,6 +211,7 @@ module RIO.Prelude.Reexports
   , Prelude.curry
   , Prelude.error
   , Prelude.even
+  , Prelude.fromEnum
   , Prelude.fromIntegral
   , Prelude.fst
   , Prelude.gcd
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -4,6 +4,7 @@
 
 * Re-export `Numeric.Natural.Natural` [#119](https://github.com/commercialhaskell/rio/issues/119)
 * Re-export `Data.Functor.<&>` from GHC 8.4+, falling back local definition for `base < 4.11` [#117](https://github.com/commercialhaskell/rio/issues/117)
+* Re-export `fromEnum` from RIO, export `toEnum`, `read` and `fromJust` from RIO.Partial
 
 ## 0.1.4.0
 


rio/src/RIO/Prelude/Reexports.hs
```diff
@@ -215,6 +216,7 @@ module RIO.Prelude.Reexports
   , Prelude.show
   , Prelude.snd
   , Prelude.subtract
+  , Prelude.toEnum
```
snoyberg: Unfortunately this function is partial, so we wouldn't want to expose it from `RIO`. As a result, I think it also makes sense to not export `fromEnum`. Perhaps a `RIO.Partial` module would make sense for these and others?

locallycompact: I'd say putting `fromEnum` in partial simply because `toEnum` is partial would feel awkward and forced. I'd say just separate them. What else could go in `RIO.Partial` that might be useful?

snoyberg: Maybe `read` would be a good fit. For that matter, we could have `head` and `tail`, but I'd imagine those would go better in `RIO.List.Partial`.

tejon:
Tossing in my pennies: `RIO.Partial` with nothing but `toEnum` and `read` doesn't bother me. I agree that `head` and `tail` belong in `RIO.List.Partial`, and also that there's really nothing wrong with providing `fromEnum` alone in base `RIO`.

rio/src/RIO/Prelude/Reexports.hs
locallycompact: What about `fromJust`?

snoyberg: That could work too, not that I'm a big fan of the function 😛

snoyberg:
@locallycompact Do you want to go ahead with the updates to this PR?

locallycompact:
How does this look?

rio/src/RIO/Partial.hs
```diff
@@ -0,0 +1,10 @@
+-- | Partial functions. Import as:
+--
+-- > import qualified RIO.Partial as P
```
snoyberg: I don't think this requires a qualified import mention, since the names don't conflict. But _if_ we were to recommend one, I'd go for the full word `Partial`; we should make this as painful as possible 😛

rio/package.yaml
```diff
@@ -54,6 +54,7 @@ library:
     - RIO.Prelude.Reexports
     - RIO.Prelude.Renames
     - RIO.Prelude.RIO
+    - RIO.Prelude.Partial
```
snoyberg: Can you add a `ChangeLog.md` entry for the changes made here?

rio/src/RIO/Prelude/Partial.hs
```diff
@@ -0,0 +1,8 @@
+module RIO.Prelude.Partial
```
snoyberg: Sorry I missed this last time around, but we don't need this layer of indirection. The RIO.Prelude modules are intended for this exported by RIO itself

locallycompact:
This ok?

snoyberg:
Perfect!

PR #122, opened 2018-07-15
Reexport Numeric.Natural.Natural

### LINKED ISSUE #119: Re-export Numeric.Natural.Natural
`Data.Int.Int{16.32.64}` are re-exported, could we have `Natural` too?

ocharles:
Fixes #119


snoyberg:
Thanks!

PR #129, opened 2018-08-21
Improve documentation of logOptionsHandle

roman:
The improved documentation explains what are all the side-effects from using a `True` value in the Debug flag

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -292,11 +292,19 @@ logOptionsMemory = do
 -- to perform verbose logging or not. Individiual settings can be
 -- overridden using appropriate @set@ functions.
 --
+-- When Verbose Flag is @True@, the following happens:
+--
+--     * @setLogVerboseFormat@ is called with @True@
+--     * @setLogUseColor@ is called with @True@
+--     * @setLogUseLoc@ is called with @True@
+--     * @setLogUseTime@ is called with @True@
+--     * @setLogMinLevel@ is called with 'Debug' log level
+--
 -- @since 0.0.0.0
 logOptionsHandle
   :: MonadIO m
   => Handle
-  -> Bool -- ^ verbose?
+  -> Bool -- ^ Verbose Flag
   -> m LogOptions
 logOptionsHandle handle' verbose = liftIO $ do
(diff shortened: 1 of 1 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -292,11 +292,19 @@ logOptionsMemory = do
 -- to perform verbose logging or not. Individiual settings can be
 -- overridden using appropriate @set@ functions.
 --
+-- When Debug Flag is @True@:
+--
+--     * enables @logVerboseFormat@ option
```
lehins: @roman Those are internal names, so might not be too informative to those who are new to the library. Because of this, do you think it could be better to reword them in such a way that the links to setters are also generated. Something along the lines:
```
-- When @Debug@ Flag is `True`, will modify all the options below:
--
--     * enables @`setLogVerboseFormat` `True`@
--     * enables @`setLogUseColor` `True`@, but only when the `Handle` is a terminal
--     * enables @`setLogUseLoc` `True`@
--     * enables @`setLogUseTime` `True`@
--     * sets @`setLogMinLevel` `Debug`@
```

lehins:
IMHO "Debug" sounds a bit overused, just like "Verbose" is. So maybe going with something like "Voluble" or "Expressive". That's just an opinion though, so whatever works, even with the proposed changes it becomes a lot more clear what's going on.

rio/src/RIO/Prelude/Logger.hs
```diff
@@ -292,11 +292,19 @@ logOptionsMemory = do
...
+--     * enables @logUseLoc@ option
+--     * enables @logUseTime@ option
+--     * sets @logMinLevel@ to 'Debug'
+--
 -- @since 0.0.0.0
 logOptionsHandle
   :: MonadIO m
   => Handle
-  -> Bool -- ^ verbose?
+  -> Bool -- ^ Debug Flag:
```
snoyberg: I don't see it as better to call this "debug flag": that implies that the only difference is whether it's showing Debug logging.

roman:
@lehins @snoyberg updated the documentation with your suggestions, can you please check it out again?

lehins:
LGTM as well. It's certainly a lot more informative than before.

PR #133, opened 2018-09-26
logUseColor now defaults to False on Windows, even when verbose

### LINKED ISSUE #131: Logger produces bad output on Windows
Hi there.
The RIO logger, used in stack, produces bad output on windows cmd, which doesn't support the bash escape sequences.
winconsole
In `Logger.replaceUnicode`, 
```haskell
simpleLogFunc lo cs _src level msg =
    when (level >= logMinLevel lo) $ do
      timestamp <- getTimestamp
      logSend lo $ getUtf8Builder $
        timestamp <>
        getLevel <>
        ansi reset <>
        msg <>
        getLoc <>
        ansi reset <>
        "\n"
  where
   reset = "\ESC[0m"
   setBlack = "\ESC[90m"
   setGreen = "\ESC[32m"
   setBlue = "\ESC[34m"
   setYellow = "\ESC[33m"
   setRed = "\ESC[31m"
   setMagenta = "\ESC[35m"
```
The function just outputs the reset sequences unconditionally.

RIO (and transitively stack) should instead use a platform independent console library such as `ansi-terminal`

rkoeninger:
Fixes #131 

Simple change in `logOptionsHandle`. Updated the haddock comment. Updated the change log.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 module RIO.Prelude.Logger
@@ -68,7 +69,6 @@ import qualified Data.ByteString as B
 import           System.IO                  (localeEncoding)
 import           GHC.Foreign                (peekCString, withCString)
 import Data.Semigroup (Semigroup (..))
-import System.Info (os)
 
 -- | The log level of a message.
 --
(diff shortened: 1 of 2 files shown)


rkoeninger:
One of the required macos builds failing with:

```
(ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 6, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "s3.amazonaws.com", service name: Just "443"): does not exist (nodename nor servname provided, or not known))
```

Does that count against this PR? I don't know what to do about it.

rio/src/RIO/Prelude/Logger.hs
```diff
@@ -315,7 +316,7 @@ logOptionsHandle handle' verbose = liftIO $ do
     , logVerboseFormat = return verbose
     , logTerminal = terminal
     , logUseTime = verbose
-    , logUseColor = verbose && terminal
+    , logUseColor = verbose && terminal && os /= "mingw32"
```
snoyberg: I'd rather use the `WINDOWS` CPP flag, so that we use the same logic throughout the codebase for checking for Windows. Otherwise, looks great!

snoyberg:
Thanks!

PR #137, opened 2018-11-02
Fix backslash to be appeared in haddock

arowM:
In the [example code](http://hackage.haskell.org/package/rio-0.1.5.0/docs/RIO.html#v:withLogFunc) for `withLogFunc` on haddock, backslash before `lf ->` is not properly rendered.
This PR just fixes it.


snoyberg:
Thanks!

PR #138, opened 2018-11-07
WIP: Add Durable Atomic File Writes to RIO

roman:


roman:
I added tests for the happy paths as those are easy to do. I also created a program that uses these functions and was able to assert via `strace` that calls to the C API were being done appropriately.

Doing automated tests for this API is rather difficult given that we cannot virtualize VMs that we can shutdown to assert fsync durability in automated CI

roman:
Hey @snoyberg when you have some time, can you do a review of this ticket, please? Thanks!

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
+{-# LANGUAGE CPP                      #-}
+{-# LANGUAGE ForeignFunctionInterface #-}
+{-# LANGUAGE NoImplicitPrelude        #-}
+{-# LANGUAGE OverloadedStrings        #-}
+module RIO.File
```
snoyberg: Please include a `@since` comment here

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+-- NOTE: Not Async exception safe
+openFileFromDir :: (MonadIO m) => Fd -> FilePath -> IOMode -> m Handle
+openFileFromDir (Fd dirFd) fp iomode =
+  liftIO $
+  withFilePath fp $ \f -> do
+    fileFd <-
+      throwErrnoIfMinus1Retry "openFileFromDir" $
+      c_safe_openat dirFd f (ioModeToFlags iomode) 0o666
+    (fD, fd_type) <-
+      FD.mkFD
```
snoyberg: Instead of having to remember to close the `fd` on exceptions here and in the other calls, what about using `bracketOnException`?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+        fileFd
+        iomode
+        Nothing {- no stat -}
+        False {- not a socket -}
+        False {- non_blocking --}
+         -- TODO: We want to investigate if we can do non-blocking here
+       `catchAny` \e -> do
```
snoyberg: I'd use `onException` here instead, unless there's a specific reason you want to ignore async exceptions here.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+        void (c_close fileFd)
+        throwIO e
+    -- we want to truncate() if this is an open in WriteMode, but only if the
+    -- target is a RegularFile. ftruncate() fails on special files like
+    -- /dev/null.
+    when (iomode == WriteMode && fd_type == RegularFile) $ Device.setSize fD 0
```
snoyberg: Do we know for a fact that this call cannot throw an exception?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+-- use later on to fsync the directory with.
+--
+-- @since 0.1.6
+openFileDurable :: MonadIO m => FilePath -> IOMode -> m (Fd, Handle)
+openFileDurable absFp iomode =  do
+  let dir = takeDirectory absFp
+      fp = takeFileName absFp
+
+  dirFd <- openDir dir
+  fileHandle <- openFileFromDir dirFd fp iomode
```
snoyberg: Looks like you need an `` `onException` close dirFd``

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+    when (iomode == WriteMode && fd_type == RegularFile) $ Device.setSize fD 0
+    HandleFD.mkHandleFromFD fD fd_type fp iomode False Nothing `onException`
+      (liftIO $ Device.close fD)
+
+-- | Opens a file using the openat C low-level API. This approach allows us to
+-- get a file descriptor for the directory that contains the file, which we can
+-- use later on to fsync the directory with.
+--
+-- @since 0.1.6
+openFileDurable :: MonadIO m => FilePath -> IOMode -> m (Fd, Handle)
```
snoyberg: What makes this function "durable"? Isn't it really `openFileAndDir`?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+  let dir = takeDirectory absFp
+      fp = takeFileName absFp
+
+  dirFd <- openDir dir
+  fileHandle <- openFileFromDir dirFd fp iomode
+  return (dirFd, fileHandle)
+
+-- | This sub-routine does the following tasks:
+--
+-- * It calls fsync and then closes the given handler (mapping to a temporal/backup filepath)
```
snoyberg:
```suggestion
-- * It calls fsync and then closes the given Handle (mapping to a temporal/backup filepath)
```

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+-- These steps guarante that the file changes are durable.
+--
+-- @since 0.1.6
+closeFileDurable :: MonadIO m => Fd -> Handle -> m ()
+closeFileDurable (Fd dirFd) h =
+  liftIO $
+  finally
+    (do fileFd <- handleToFd h
+        fsyncDescriptor "File" (FD.fdFD fileFd) `finally` hClose h
+        fsyncDescriptor "Directory" dirFd)
```
snoyberg: To confirm: you want the logic "if `fsyncDescriptor` on the file fails, we shouldn't even try to do it on the directory. Is that correct?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+    fsyncDescriptor name fd =
+      void $
+      throwErrnoIfMinus1 ("closeFileDurable - fsync(" <> name <> ")") $
+      c_safe_fsync fd
+    closeDirectory =
+      void $
+      throwErrnoIfMinus1Retry "closeFileDurable - close(Directory)" $
+      c_close dirFd
+
+toTmpFilePath :: FilePath -> FilePath
```
snoyberg: I don't think this is sufficient, instead you should use the template file name functions from `System.IO` to ensure the file doesn't already exist.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+        Just fd -> return fd
+        Nothing -> error "not a file handle"
+    HandleFD.DuplexHandle {} -> error "not a file handle"
+# else
+handleToFd = HandleFD.handleToFd
+#endif
+
+-- | This sub-routine does the following tasks:
+--
+-- * It calls fsync and then closes the given handler (mapping to a temporal/backup filepath)
```
snoyberg:
```suggestion
-- * It calls fsync and then closes the given Handle (mapping to a temporal/backup filepath)
```

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+-- * It renames the file to the original path (using renameat)
+-- * It calls fsync and then closes the containing directory of the file
+--
+-- These steps guarante that the file are durable, and that the backup mechanism
```
snoyberg:
```suggestion
-- These steps guarantee that the file is durable, and that the backup mechanism
```

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+-- for catastrophic failure is discarded after no error is thrown.
+--
+-- @since 0.1.6
+closeFileDurableAtomic ::
+     MonadIO m => FilePath -> Fd -> Handle -> m ()
+closeFileDurableAtomic filePath (Fd dirFd) fileHandler =
```
snoyberg: There's a lot of overlapping logic with the above. Since this code is so sensitive to proper exception handling, I think refactoring to share the code would be warranted.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+#endif
+
+-- | After a file is closed, it opens it again and executes fsync internally on
+-- both the file and the directory that contains it.
+--
+-- [The effectiveness of calling this function is
+-- debatable](https://stackoverflow.com/questions/37288453/calling-fsync2-after-close2/50158433#50158433),
+-- as this relies on internal implementation details at the Kernel level that
+-- might change. We argue that, despite this fact, calling this function may
+-- bring benefits in terms of durability
```
snoyberg: Is this intended to be used together with the other durable functions in this module, or with non-durable file writing functions?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+--
+-- * It reliably saves the file in the correct directory; including edge case
+--   situations like a different device being mounted to the current directory,
+--   or the current directory being renamed to some other name while the file is
+--   being used.
+--
+-- * It ensures durability by executing fsync before close
+--
+-- @since 0.1.6
+withFileDurable ::
```
snoyberg: Should this instead be `withFileBinaryDurable`? We tend to prefer binary file handles everywhere else.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+     MonadUnliftIO m => FilePath -> IOMode -> (Handle -> m r) -> m r
+withFileDurable absFp iomode cb =
+#if WINDOWS
+  withFile absFp iomode cb
+#else
+  withRunInIO $ \run ->
+    bracket
+      (openFileDurable absFp iomode)
+      (uncurry closeFileDurable)
+      (\(_, h) -> do run (cb h))
```
snoyberg:
```suggestion
      (run . cb . snd)
```

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,345 @@
...
+--
+-- * It keeps all changes in a copy file, and after is closed it renames it to
+--   the original filepath, in case of catastrophic failure, the original file
+--   stays unaffected.
+--
+--
+-- === Performance Considerations
+--
+-- When using an 'IOMode' of 'WriteMode', 'ReadWriteMode', or 'AppendMode', this function
+-- performs a copy operation of the file being read. This may be prohibitive in
```
snoyberg: I don't understand why we need to do a copy here instead of doing an atomic move.

rio/src/RIO/File.hs
roman: Nice recommendation... I moved the creation of the `fileFd` variable to a `brackeOnError`

rio/src/RIO/File.hs
roman: Good catch, given it's IO nature and it is a typeclass function, I wouldn't rely on that being the case. I added this section into a `bracketOnError` (suggested in a previous comment). I think that should cover our bases, do you agree?

rio/src/RIO/File.hs
roman: Yes, that was my intent, would you see any benefits around executing the fsync on the directory?

rio/src/RIO/File.hs
roman: Uhm, you are right, unfortunately, the `System.IO` does not export the functionality to generate only the `FilePath` but rather it opens a `Handle` at the same time, I'll have to either use the existing function and close the file handle or re-implement the file name generation function in this module.

rio/src/RIO/File.hs
roman: Is intended to be used with non-durable file writing functions. The idea is if you have a call in your module that doesn't enforce durability (fsync), calling this function after that would at least attempt to fsync after it was closed the first time. There is no 100% guarantee this will work, but in some systems, it may...

rio/src/RIO/File.hs
roman: These names are getting pretty _Germanic_ 🙈

rio/src/RIO/File.hs
roman: Well, my thoughts were, in case there is a hard restart in the middle of the write operation, the original file should stay the same, by moving the file, the guarantee is not there. This is a very expensive operation though

roman:
@snoyberg when you have a second, can you take a second look, please? thanks in advance.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,349 @@
...
+             Nothing {- no stat -}
+             False {- not a socket -}
+             False {- non_blocking --}
+         -- we want to truncate() if this is an open in WriteMode, but only if the
+         -- target is a RegularFile. ftruncate() fails on special files like
+         -- /dev/null.
+         when (iomode == WriteMode && fd_type == RegularFile) $
+           Device.setSize fD 0
+         HandleFD.mkHandleFromFD fD fd_type fp iomode False Nothing
+           `onException` (liftIO $ Device.close fD))
```
snoyberg: I'm not sure about this bit. We already have a `c_close` error handling installed via the `bracketOnError`. Won't we end up with a double close here via the `Device.close` call? And if not, why don't we need a `onException` bit for the `setSize` call?

rio/src/RIO/File.hs
snoyberg: Yes, though I have some new concerns. I'll start as a new review.

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,349 @@
...
+-- * It renames the file to the original path (using renameat)
+-- * It calls fsync and then closes the containing directory of the file
+--
+-- These steps guarantee that the file is durable, and that the backup mechanism
+-- for catastrophic failure is discarded after no error is thrown.
+--
+-- @since 0.1.6
+closeFileDurableAtomic ::
+     MonadUnliftIO m => FilePath -> FilePath -> Fd -> Handle -> m ()
+closeFileDurableAtomic tmpFilePath filePath dirFd@(Fd cDirFd) fileHandler = do
```
snoyberg: s/fileHandler/fileHandle

rio/src/RIO/File.hs
snoyberg: No objection, but please clarify that this is intended to work around non-durability in other APIs, as opposed to being necessary for the API functions provided here.

rio/src/RIO/File.hs
snoyberg: Ugh, I messed up, it _should_ be `withBinaryFileDurable` for consistency, sorry.

rio/src/RIO/File.hs
snoyberg: I'm still not following, sorry. The typical approach for atomic file writing is to write the full contents to a temp file in the same directory and then rename the file atomically. I don't see where a file copy protects against a hard restart in the middle of a write operation.

roman: Ok, it is likely I'm missing something here, so you tell me how a move would work in the following example:

I have a `db.txt` file on my program that I want to open in `ReadWriteMode`, and I use a `withBinaryFileDurableAtomic` call, I do some reads on the file and then some other writes:

```haskell
withBinaryFileDurableAtomic "db.txt" ReadWriteMode $ \h ->
  -- some reads (1)
  -- some writes start (2)
  -- calculations that take considerate time (3)
  -- some writes (4)
```

Under the assumption that the system restarts/crashes in step (3), I understand that if we 

(a) do a copy of `db.txt` to  a temp file `.db.txt123abc.swp` and rename back after successful writes are done

both files, an unmodified `db.txt` and a temp file copy `.db.txt123abc.swp` files are going to be present in the file system. If we run the algorithm again, is going to start as if `db.txt` was never modified.

(b) do a move/rename of `db.txt` to `.db.txt123abc.swp` and rename back after successful writes are done

the original file `db.txt` is not going to be present at the restart, only the unmodified `.db.txt123abc.swp` file, which would make the original algorithm not work.

I'm now noticing you are mentioning an _atomic_ move, which maybe it means something different from a regular `mv`/`rename` call.

Apologies for not making a better effort at explaining my thoughts before and thank you for helping me figure this out and spend some minutes going through this.

roman:
>  For future revisions, can you please avoid any rebasing of the commits?

Oh apologies about that 🙈 , yes I will avoid rebasing up until the point the pull request is completely reviewed. Not sure if you care too much about the commit history being pristine (we don't want people to ever know we make mistakes _ever_, am I right 😜?).

rio/src/RIO/File.hs
roman: Argh, great catch, I'll refactor this so that the `mkFD` is part of the open subroutine of the `bracket` call

snoyberg: Nice setup. On a separate note, we may want to save this code as an example of where Rust is far easier to write correctly than Haskell.

rio/src/RIO/File.hs
snoyberg: Atomic move is the same as `mv`/`rename`, sorry for the confusion.

I think the issue here is that I wasn't thinking of `ReadWriteMode`, but rather `WriteMode`. For `WriteMode`, I'm fairly certain the copy isn't necessary, right? It's really only a question for `ReadWriteMode`. I'm not even convinced there that we want to copy, as opposed to moving to a temp file location and potentially losing the file on a crash.

roman: > I think the issue here is that I wasn't thinking of ReadWriteMode, but rather WriteMode. For WriteMode, I'm fairly certain the copy isn't necessary, right?

I'm not sure, I would say it depends if the file is open _only_ as a `WriteMode` on your program lifecycle, in this situation, this statement makes sense. So in this case, you want something that is atomic but not durable (you don't care about the previous save).

> I'm not even convinced there that we want to copy, as opposed to moving to a temp file location and potentially losing the file on a crash.

Well, I argue losing the given file on a crash should be a decision made by the caller of this API, not by us. IMO, there is no atomicity if you can lose the file on a crash, also, there is already a function that offers durability without atomicity that the caller can use if they don't want to incur on the `copy` costs.

On a different note, I updated the documentation of the module with more information as a recommendation from @nh2

roman:
ping @snoyberg

roman:
@snoyberg all discussions were resolved, can you review the new changes after CI passes? thank you!

rio/src/RIO/File.hs
```diff
@@ -40,9 +47,12 @@ For this use case, you want to use the functions:
...
 * 'writeBinaryFileDurable'
 
-The regular use case for this scenario happens when your program deals with the
-atomicity of your data through other means. (e.g. SQLite Write Ahead Log (WAL)
-strategy).
+The regular use case for this scenario happens when your program deals with file
+modifications that must be guaranteed to be durable, but you don't care that
+changes are consistent. If you use this function, more than likely your program
+is ensuring consistency guarantees through other means, for example, SQLite uses
+the (Write Ahead Log) WAL algorithm to ensure changes are atomic at an
```
snoyberg:
```suggestion
 the Write Ahead Log (WAL) algorithm to ensure changes are atomic at an 
```

rio/src/RIO/File.hs
```diff
@@ -7,31 +7,38 @@
...
+file gets corrupted or lost at one particular execution of your program given
+that is cheap to execute the data export program a second time. In other words,
+your program doesn't /rely/ on the data contained in this file in order to work.
 
 === Atomic but not Durable
 
-The regular use case for this scenario is when you have a multi-step algorithm
-that builds temporal artifacts, you care about them being atomic, but your
-program recreates them each time. (e.g. @ghc --make@ generating @.o@ files).
+Imagine an scenario where your program builds a temporary file that serves as an
```
snoyberg:
```suggestion
 Imagine a scenario where your program builds a temporary file that serves as an 
```

rio/src/RIO/File.hs
```diff
@@ -147,27 +165,46 @@ openDir fp
...
     c_safe_fsync cFd
 
--- NOTE: Not Async exception safe
+-- | Opens a file from a directory, using this function in favour of a regular
+-- 'openFile' guarantees that any file modifications are kept in the same
+-- directory where the file was opened. An edge case scenario is a mount
+-- happening in the directory where the file was opened while your program is
+-- running.
+--
+-- If you use this function, make sure you are working on an unmasked state,
```
snoyberg: I think you mean masked state.

rio/src/RIO/File.hs
```diff
@@ -382,36 +419,42 @@ withBinaryFileDurable absFp iomode cb =
 -- * It ensures durability by executing an fsync call before closing the file
 --   handle
 --
--- * It keeps all changes in a copy file, and after is closed it renames it to
---   the original filepath, in case of catastrophic failure, the original file
---   stays unaffected.
+-- * It keeps all changes in a temporal file, and after is closed it atomically
```
snoyberg:
```suggestion
 -- * It keeps all changes in a temporary file, and after it is closed it atomically 
```

rio/src/RIO/File.hs
```diff
@@ -382,36 +419,42 @@ withBinaryFileDurable absFp iomode cb =
...
 --
 --
 -- === Performance Considerations
 --
--- When using a non read-only 'IOMode' (e.g. 'WriteMode', 'ReadWriteMode',
--- 'AppendMode'), this function performs a copy operation of the specified input
--- file to guarantee the original file is intact in case of a catastrophic
--- failure (no partial writes). This approach may be prohibitive in scenarios
--- where the input file is expected to be large in size.
+-- When using a read-write 'IOMode' (e.g. 'ReadWriteMode', 'AppendMode'), this
```
snoyberg:
```suggestion
 -- When using a writable but non-truncating 'IOMode' (i.e. 'ReadWriteMode' and 'AppendMode'), this 
```

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,428 @@
...
+  liftIO $
+  withFilePath fp $ \cFp ->
+    Fd <$>
+    (throwErrnoIfMinus1Retry "openDir" $
+     c_safe_open cFp (ioModeToFlags ReadMode) 0o660)
+
+closeDirectory :: MonadIO m => Fd -> m ()
+closeDirectory (Fd dirFd) =
+  liftIO $
+  void $
```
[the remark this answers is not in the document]
roman: > Are the only possibilities -1 and 0?

Correct, that's what the documentation (man page) says. The `-1` is being dealt with `throwErrnoIfMinus1Retry`

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,428 @@
...
+    dirPath = takeDirectory filePath
+    filename = takeFileName filePath
+    tmpFilename = "." <> filename <> ".tmp"
+
+handleToFd :: Handle -> IO FD.FD
+#if MIN_VERSION_base(4,9,0)
+handleToFd h =
+  case h of
+    HandleFD.FileHandle _ mv -> do
+      HandleFD.Handle__{HandleFD.haDevice = dev} <- readMVar mv
```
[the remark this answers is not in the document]
roman: Not opposing to the suggestion, but I'm very curious to learn what "funny business" entails here, are we concerned another thread might `takeMVar` before-hand and block the call here?

How would the usage of `withMVar` prevent this scenario? Looking at it's implementation:

```
withMVar :: MVar a -> (a -> IO b) -> IO b
withMVar m io =
  mask $ \restore -> do
    a <- takeMVar m
    b <- restore (io a) `onException` putMVar m a
    putMVar m a
    return b
```

Wouldn't a call to `takeMVar` block our thread the same in such scenario?

rio/src/RIO/File.hs
```diff
@@ -0,0 +1,428 @@
...
+          (hClose . snd)
+          (return . fst)
+
+toTmpFilePath :: MonadUnliftIO m => FilePath -> m FilePath
+toTmpFilePath filePath =
+    buildTemporaryFilePath (dirPath </> tmpFilename)
+  where
+    dirPath = takeDirectory filePath
+    filename = takeFileName filePath
+    tmpFilename = "." <> filename <> ".tmp"
```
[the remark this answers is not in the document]
roman: I thought this was good giving I'm relying on the `openBinaryTempFile` call on line 262 to ensure uniqueness. 

I took a look [at its implementation](http://hackage.haskell.org/package/base-4.12.0.0/docs/src/System.IO.html#openTempFile'), and I'm noticing it will loop until it finds a unique name. Even in the situation of a race condition, the original algorithm generates a random String for each attempt (it keeps a global IORef counter to change a suffix number on each call).

roman:
Hey @snoyberg, I covered all the new set of comments/suggestions, thanks for taking the time to do a great review.

rio/src/RIO/File.hs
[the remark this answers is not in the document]
snoyberg: I'm worried that you can call `readMVar` and then, while using the contents of the `MVar`, another thread could still access the `MVar`. I'd rather block any other threads while we're using those contents.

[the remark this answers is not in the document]
roman: Uhmm, I think other APIs that make use of that MVar do not like me locking that var, I get these errors on the test-suite:

```
test/RIO/FileSpec.hs:37:
      1) RIO.File.withBinaryFileDurableAtomic, happy path, works the same as withFile
           uncaught exception: IOException of type ResourceBusy (/tmp/rio16337/with_file_durable_atomic: openBinaryFile: resource busy (file is locked))

      test/RIO/FileSpec.hs:47:
      2) RIO.File.withBinaryFileDurable, happy path, works the same as withFile
           uncaught exception: IOException of type ResourceBusy (/tmp/rio16337/with_file_durable: openBinaryFile: resource busy (file is locked))
```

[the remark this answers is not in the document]
roman: Also something I think is worth mentioning, This is the implementation from [what is currently in `base`](http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Handle.FD.html#handleToFd)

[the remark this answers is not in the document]
roman: You are 100% right, I should have improved upon this from the getgo, and especially if you are pointing it out — apologies for that. I did changes that should work now with a `withMVar` invocation. Can you take a look? thank you 👍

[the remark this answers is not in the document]
snoyberg: No worries. This looks good to me now, thanks!

snoyberg:
@roman I believe that addresses all concerns, should we merge this now?

snoyberg:
Thanks!

PR #152, opened 2019-02-27
Add a Deque datatype (fixes #151)

### LINKED ISSUE #151: Include some mutable containers
In particular, `Deque` and possibly `DLList` from [`Data.Mutable`](https://www.stackage.org/haddock/lts-13.9/mutable-containers-0.3.4/Data-Mutable.html#g:4). I have a specific need for a `Deque` (or as list single-ended queue) in a rewrite of a `Writer` monad in yesod-form.

CC @lehins @nh2, I think you may have some feedback.

snoyberg:
CC @lehins


rio/src/RIO/Deque.hs
```diff
@@ -0,0 +1,258 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE TypeFamilies #-}
```
lehins: I might be missing it, but it doesn't look like `TypeFamilies` is necessary here anymore, same with `DequeSpec.hs` file

rio/src/RIO/Deque.hs
```diff
@@ -0,0 +1,258 @@
...
+import           Control.Exception            (assert)
+import           Control.Monad                (liftM)
+import qualified Data.Vector.Generic.Mutable  as V
+import qualified Data.Vector.Mutable          as B
+import qualified Data.Vector.Storable.Mutable as S
+import qualified Data.Vector.Unboxed.Mutable  as U
+import           Data.Primitive.MutVar
+
+data DequeState v s a = DequeState
+    (v s a)
```
lehins: Out of curiosity, why is the vector field not strict? I noticed it in the original code too, is it better for some reason to keep it lazy?

snoyberg: It's an interesting question. Firstly, the primary performance benefit of making it strict (unpacking) wouldn't be possible, since it's a polymorphic field. The correctness argument of strict fields (avoiding bottom values and unevaluated values) would apply, and is a plus. There _may_ be a small performance impact though from making it strict. I only heard it offhand once, but apparently there's an extra is-evaluated check adding to all usages of the data constructor. Overall: I'd prefer to make it strict for correctness, and probably didn't have a good reason to avoid it in the first place.

PR #154, opened 2019-03-03
Add `getDequeSize`, `dequeToVector`, `freezeDeque`

lehins:


rio/src/RIO/Deque.hs
```diff
@@ -238,6 +254,19 @@ dequeToList
 dequeToList = foldrDeque (\a rest -> pure $ a : rest) []
 {-# INLINE dequeToList #-}
 
+
+-- | Convert to an immutable vector of any type
```
snoyberg: It would be good to have some docs explaining the difference between `dequeToVector` vs `freezeDeque` a bit more explicitly.

lehins: Done. Added examples as well.

PR #155, opened 2019-03-04
Move incorrectly classified Map conversion functions: `toAscList`, `toDescList`

lehins:
I don't think there is anything wrong/unchecked with:
* `Data.Map.Strict.toAscList`
* `Data.Map.Strict.toDescList`

They do not have any preconditions, other than the `Map` itself be valid. In fact `RIO.Set` exports corresponding functions.

I did not remove them from `RIO.Map.Unchecked`, otherwise it would be a breaking change, so it should be done at the next major release instead.


lehins:
Done

PR #157, opened 2019-03-15
add lifted versions of `System.Exit.exitFailure` and `exitSuccess`

ketzacoatl:
into the Rio `Prelude`, and re-exported from there.

### DIFF
--- a/rio/src/RIO.hs
+++ b/rio/src/RIO.hs
@@ -1,5 +1,6 @@
 module RIO
   ( module RIO.Prelude.Display
+  , module RIO.Prelude.Exit
   , module RIO.Prelude.Extra
   , module RIO.Prelude.IO
   , module RIO.Prelude.Lens
@@ -15,6 +16,7 @@ module RIO
   ) where
 
 import RIO.Prelude.Display
+import RIO.Prelude.Exit
 import RIO.Prelude.Extra
 import RIO.Prelude.IO
 import RIO.Prelude.Lens
--- a/rio/src/RIO/Prelude/Exit.hs
+++ b/rio/src/RIO/Prelude/Exit.hs
@@ -1,0 +1,31 @@
+module RIO.Prelude.Exit
+  ( exitFailure
+  , exitSuccess
+  , exitWith
+  , System.Exit.ExitCode(..)
+  ) where
+
+import           Control.Monad.IO.Class
+import qualified System.Exit ( ExitCode (..)
+                             , exitFailure
+                             , exitSuccess
+                             , exitWith
+                             )
+
+-- | Lifted version of "System.Exit.exitFailure".
+--
+-- @since 0.1.9.0.
+exitFailure :: MonadIO m => m a
+exitFailure = liftIO System.Exit.exitFailure
+
+-- | Lifted version of "System.Exit.exitSuccess".
+--
+-- @since 0.1.9.0.
+exitSuccess :: MonadIO m => m a
+exitSuccess = liftIO System.Exit.exitSuccess
+
+-- | Lifted version of "System.Exit.exitWith".
+--
+-- @since 0.1.9.0.
+exitWith :: MonadIO m => System.Exit.ExitCode -> m a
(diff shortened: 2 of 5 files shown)


rio/src/RIO/Prelude/Exit.hs
```diff
@@ -0,0 +1,15 @@
+module RIO.Prelude.Exit
+  ( module System.Exit
```
snoyberg: I prefer having explicit export lists to avoid accidentally leaking functionality added in later versions of base. Can you either:

1. Make this list explicit
2. Make the import list below explicit

rio/src/RIO/Prelude/Exit.hs
```diff
@@ -0,0 +1,15 @@
...
+  , exitSuccess
+  , exitFailure
+  ) where
+
+import           Control.Monad.IO.Class
+import           System.Exit hiding( exitFailure, exitSuccess )
+import qualified System.Exit
+
+exitFailure :: MonadIO m => m a
```
snoyberg: Please add Haddock comments that include a message like `Lifted version of "System.Exit.exitFailure"` and a `@since 0.1.9.0` comment.

rio/src/RIO/Prelude/Exit.hs
```diff
@@ -0,0 +1,15 @@
...
+exitFailure = liftIO System.Exit.exitFailure
+
+exitSuccess :: MonadIO m => m a
+exitSuccess = liftIO System.Exit.exitSuccess
```
snoyberg: Should we also include `exitWith`? And I think `ExitCode` is already reexported elsewhere, but perhaps for logical consistency it should be reexported here.

rio/src/RIO.hs
```diff
@@ -1,5 +1,6 @@
 module RIO
   ( module RIO.Prelude.Display
+  , module RIO.Prelude.Exit
```
snoyberg: Please include a mention in the `ChangeLog.md` about the addition.

ketzacoatl:
Thanks for the great feedback, I will make those changes and update the PR.

rio/src/RIO/Prelude/Exit.hs
ketzacoatl: I'd agree, done!

rio/src/RIO/Prelude/Exit.hs
ketzacoatl: Definitely, thanks!

rio/src/RIO.hs
ketzacoatl: This would end up in a version bump to `0.1.9.1`? or does the new entry go under `0.1.9.0`? or does this go under a "next release" heading in the `ChangeLog`?

snoyberg: Since I haven't released 0.1.9.0 yet, it should be part of that release.

rio/src/RIO/Prelude/Exit.hs
```diff
@@ -0,0 +1,28 @@
...
+
+import           Control.Monad.IO.Class
+import qualified System.Exit ( ExitCode (..)
+                             , exitFailure
+                             , exitSuccess
+                             , exitWith
+                             )
+
+--
+-- | Lifted version of "System.Exit.exitFailure", @since 0.1.9.0.
```
snoyberg: This `@since` bits should be on their own lines to get them to render correctly.

rio/src/RIO.hs
snoyberg: Also, this new module should be added to the hidden module list in `package.yaml`

ketzacoatl: I'm not sure about the hidden module list, but I'm assuming you mean the existing `other-modules`. I've updated that, thanks!

rio/src/RIO/Prelude/Exit.hs
ketzacoatl: OK, updated.

Issue #161, opened 2019-04-03
Haddock documentation layout improvements


brendanhay:
When viewing the main entry-point to RIO on [Hackage/Stackage](http://hackage.haskell.org/package/rio/docs/RIO.html), you see a **Documentation** heading and then you're straight into `Utf8Builder`, `Display`, etc. Looking at the right-side contents menu, we see **Standard logging functions**, **Run with logging**, and so on:

![Screenshot from 2019-04-03 09-47-30](https://user-images.githubusercontent.com/4493/55461932-f5c22280-55f5-11e9-83bd-96ee7169e7f0.png)

Now, If I was perhaps looking for the penultimate `RIO` newtype - I would peruse the contents menu and see .. nothing? Hmm.

So Ctrl-F still seems to work in Netscape Navigator and searching for "newtype RIO" yields the following:

![Screenshot from 2019-04-03 09-52-01](https://user-images.githubusercontent.com/4493/55462025-286c1b00-55f6-11e9-9daf-df9e34612a74.png)

_Wat._

What follows is highly personal preference/opinion/bias, but hopefully constructive:

1. Even if your project has a great `README`, which RIO most certainly does, it's still nice to have a brief introduction full of linked-Hackage identifiers I can click on to get places quickly in the heading of the `RIO` module documentation. Doesn't need to be in every leaf module, probably just `RIO`.

2. The right contents menu needs more focus on high-level concepts as main headings, with sub-headings to drill down as deemed necessary/useful. Imagine mirroring how a dead-tree technical reference organises its table of contents.

Example:

* The RIO Monad (not-a-transformer)
* Containers
* State, Writer, and Mutable References
* Logging
  - Options
  - Advanced Logging
  - Textual Output
* Displaying Datatypes

There's alot of top-level headings such as **Text**, and **Display** that then only group logging related functions - the `Display` type class is not under its respective heading.

3. All `whenM`, `nubOrd`, `view`, etc. functions are just sort of 'exported'. I tend to still think of things in terms of their 'classic' namespaces - ie. `Control` -> `Control.Monad` -> `when/M` - I understand people new to Haskell probably don't have this baggage, but still grouping things similarly in terms of headings/exports lets people see what the analogies are with pre-existing `base`, preludes, and so on. List things go under a List heading, Either under Either, writeFile and so on under Monad/Unlift/IO, and so forth.

4. Lenses. These probably deserve their own section.

[...]

PR #163, opened 2019-04-07
Documentation Improvements

lehins:
This PR contains a few things, but the major goal is documentation improvements raised in #161 

**NOTE** - It is ready to be merged (module the commit squashing), but it is WIP because I'd like to get some feedback on it.

There is a live version of the hadock generated from this PR, which is available here: https://lehins.net/rio/

Few things happed in this PR:
* Addition of two public modules:
  * `RIO.Prelude`
  * `RIO.Prelude.Types`
* Inclusion of few functions and types, see ChangeLog for those changes
* Rearrangement of `RIO` module to make it clearer.

One of the biggest issues I have with current `RIO` module and  in fact even with usual `Prelude` is that those pages are huge. They take very long time to load and it takes ages to scroll through to find anything on them. All this is due to type classes and the list of all instances for each type. It looks like the situation might get improved with future haddock version: haskell/haddock#1007 but as of now that feature is not yet available. Even with all collapsed instances by default, it will still take forever to load a page with that amount of instances on it. For RIO it is especially a problem, since it brings quite a few libraries together, it makes such common classes as `Eq`, `Show`, etc. contain a list of instances that are pages and pages long. 

The very nice solution IMHO to the above problem is to keep all of the types and type classes on their own page, while separating all of the functions in their own module. Even the functions that are part of the classes (eg. `fmap` in `Functor`) will not be exported together with the class itself. See this solution in action:
* [`RIO.Prelude`](https://lehins.net/rio/rio-0.1.10.0/RIO-Prelude.html)
* [`RIO.Prelude.Types`](https://lehins.net/rio/rio-0.1.10.0/RIO-Prelude-Types.html)

One other important (possibly breaking) change included in this PR is the `MonadFail` export. Namely the `fail` method from `Control.Monad.Monad` is no longer exported. According to the original proposal, that method will be removed in the next major version of GHC-8.6, so I though it be a good time to prepare RIO for that change.


rio/ChangeLog.md
```diff
@@ -1,5 +1,15 @@
 # Changelog for rio
 
+## 0.1.10.0
```
snoyberg: I haven't actually released 0.1.9.0 yet, so if you'd like, we can include this change there.

rio/src/RIO.hs
```diff
@@ -1,32 +1,119 @@
...
+  , module RIO.Prelude.Exit
+    -- * Mutable Variables
+    -- ** SomeRef
+  , module SomeRef
+    -- ** URef
   , module RIO.Prelude.URef
-  , module RIO.Prelude.Simple
+    -- ** IORef
+  , module UnliftIO.IORef
+    -- ** IORef
```
snoyberg: MVar?

rio/src/RIO.hs
```diff
@@ -1,32 +1,119 @@
...
+  , module UnliftIO.MVar
+#if MIN_VERSION_unliftio(0, 2, 8)
```
snoyberg: I'd be OK with a lower bound instead TBH

rio/src/RIO/Prelude/Extra.hs
```diff
@@ -75,15 +81,28 @@ nubOrd =
...
-unlessM boolM action = do
-  x <- boolM
-  if x then return () else action
+unlessM boolM action = boolM >>= (`unless` action)
+
+
+-- | Same as `(>>)`, but with arguments flipped.
+--
+-- @since 0.1.10.0
+(<<) :: Monad m => m a -> m b -> m a
```
snoyberg: Overall not so interested. Shouldn't we be encouraging `*>` and `<*` instead?

rio/src/RIO/Prelude/Extra.hs
```diff
@@ -75,15 +81,28 @@ nubOrd =
...
+(<<) = flip (>>)
+
+infixr 1 <<
+
+#if !MIN_VERSION_base(4, 11, 0)
```
snoyberg: This is a nice idea. However, it can cause naming conflicts with other libraries that do the same thing, e.g. lens. Maybe that's not so important. I've definitely wanted this operator before...

rio/src/RIO/Prelude/Renames.hs
```diff
@@ -28,8 +32,12 @@ type GVector = GVector.Vector
 type LByteString = BL.ByteString
 type LText = TL.Text
 
-toStrictBytes :: LByteString -> ByteString
```
snoyberg: I totally for we had this function!

rio/src/RIO/Prelude/Types.hs
```diff
@@ -0,0 +1,347 @@
+module RIO.Prelude.Types
```
snoyberg: I think this needs to be added to `other-modules`, right? In fact, to ensure all modules are properly listed, perhaps we should stop using hpack's auto-detection for this package.

lehins: Nah, that's the opposite of my intentions. Adding it to `other-module` will make the `RIO.Prelude` module bloated again, cause all types and classes and all their instances will move there. I want to avoid precisely that problem and for that reason kept the types in a separate module that can explored separately when one is looking for information about types, classes and instances.

lehins: I do very much dislike `hpack`s auto detection feature. Quite often it messes with my workflow.

rio/src/RIO/Prelude/Extra.hs
tejon: My sole input on this is that `>>` and `<<` should be a package deal: both, or neither. And I think the arguments presented for "both" are solid.

(I wasn't aware of the associativity difference between `<<` and `<*`... wonder what the reasoning was there?)

rio/src/RIO/Prelude/Renames.hs
tejon: Nor did I, and I'm _using_ `RIO.ByteString.Lazy.toStrict`! Docs project pays off before it's even merged. :grin:

lehins:
I think this PR also addresses #61

.azure/azure-linux-template.yml
```diff
@@ -8,7 +8,7 @@ jobs:
         BUILD: stack
       stack-lts-11:
         BUILD: stack
-        ARGS: "--resolver lts-11"
+        ARGS: "--stack-yaml stack.lts-11.yaml"
```
snoyberg: It's still a good idea, though not required, to keep the `--resolver lts-11` in addition to the `--stack-yaml`, since `--resolver lts-11` will force the latest lts-11 minor release to be used. Since there are unlikely to be any new lts-11 releases, I'm not too bothered by this.

rio/src/RIO.hs
```diff
@@ -1,32 +1,114 @@
 module RIO
-  ( module RIO.Prelude.Display
-  , module RIO.Prelude.Exit
-  , module RIO.Prelude.Extra
-  , module RIO.Prelude.IO
-  , module RIO.Prelude.Lens
+  (
+    -- * Custom @Prelude@
+
+    -- | One of the core features of @rio@ is that it can be used as @Prelude@
```
snoyberg:
```suggestion
    -- | One of the core features of @rio@ is that it can be used as a @Prelude@
```

rio/src/RIO.hs
```diff
@@ -1,32 +1,114 @@
...
+    -- replacement. Therefore it is best to disable default `Prelude` with:
```
snoyberg:
```suggestion
    -- replacement. Therefore it is best to disable the default `Prelude` with:
```

rio/src/RIO/Prelude/Extra.hs
snoyberg: > First of all they can have different implementation. As such they can have different behavior and at the very least they can have different strictness properties.

Not worried about this, that would be a bug.

> They do restrict the type to a Monad, rather then the Applicative when used in a function.

Why is that relevant?

> Last, but not least, they have different infix precedence, as such *> and <* don't compose well with >>= and =<< respectfully.

Ouch, ok.

rio/src/RIO/Prelude/Extra.hs
snoyberg: lol ok

rio/src/RIO/Prelude/Renames.hs
snoyberg: Yes, forgot. I forgot the word I was typing.

rio/src/RIO/Prelude/Types.hs
snoyberg: OK, let's stop using autodetection then, that way there's no ambiguity when merging this PR which modules are supposed to be exposed and which ones hidden.

lehins: Done

rio/src/RIO/Prelude/Extra.hs
lehins: @tejon Regret to announce, we've been outvoted. `<<` operator is now removed.

snoyberg: FTR, this is because I hadn't realized that `<*` and `<<` were intended to behave differently. The motivation makes sense in context of `Monad`, but I'm highly skeptical of creating confusing operator soup at that point.

PR #164, opened 2019-04-13
Remove accidental reexport of `Control.Applicative.empty`  

lehins:
* Remove reexport of `empty`, which was introduced in `rio-0.1.9.0`
* Bumps up version
* Updates changelog together with a missed change for the previous version

### DIFF
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -3,6 +3,7 @@
 ## 0.1.9.1
 
 * Remove accidental reexport of `Control.Applicative.empty` introduced in the previous release.
+* Functions from `Data.Data.Data` class are brought to the re-export list as well.
 
 ## 0.1.9.0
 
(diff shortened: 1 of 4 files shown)


rio/ChangeLog.md
```diff
@@ -15,6 +19,7 @@
 * Re-export `zipWith` and `runST` from `RIO.Prelude`
 * Re-export `Exception`, `MonadFail`, `Typeable` and `ST` from `RIO.Prelude.Types`
 * Switch to `MonadFail.fail` from `Monad.fail` and re-exported it from `RIO.Prelude`
+* Functions from `Data.Data.Data` class are no longer exported.
```
snoyberg: Is this an intentional change versus previous releases?

lehins: It was not intentional, but I totally support it. Generic programming has no place in Prelude IMHO. So it might have been a subconscious decision ;) Just kidding of course. I was going type, by type, class by class, and was trying to make the previous PR having least semantic changes possible, but some things slipped through.

lehins: The type class itself is exported, which is very useful for deriving.

snoyberg: I think I'd rather add them back. They haven't caused any trouble that I'm aware of, and I'd rather avoid breaking backwards compatibility if we can.

lehins: I'll humbly disagree. But it's your call, so I added them back to the reexport list

Issue #160, opened 2019-04-01
`withBinaryFileDurableAtomic` fails with nested folders


lehins:
Whenever a file that is being written to is located in a sub directory `withBinaryFileDurableAtomic` results in a runtime exception during the atomic rename.

Same problem applies to `writeBinaryFileDurableAtomic`.

Here is a short repro:

```haskell

import Import
import RIO.Directory
import RIO.File

run :: RIO App ()
run = do
  logInfo "We're inside the application!"

  let foo = "foo.txt"
  withBinaryFileDurableAtomic foo WriteMode $ \h -> do
    hPutBuilder h "foo"
  logInfo $ "Written " <> fromString foo

  let barFoo = "bar/foo.txt"
  createDirectoryIfMissing True barFoo
  withBinaryFileDurableAtomic barFoo WriteMode $ \h -> do
    hPutBuilder h "bar - foo"
  logInfo $ "Written " <> fromString barFoo
```

Which results in:

```shell
$ stack run
We're inside the application!
Written foo.txt
atomic-durable-exe: closeFileDurableAtomic - renameFile: does not exist (No such file or directory)
```

PR #167, opened 2019-04-22
Atomic writes

lehins:
This PR implements atomic file writes and fixes a bunch of issues with durable atomic and only durable file writes. In order to see the actual locations with short description of each issue [see this PR](https://github.com/lehins/rio/pull/2/files) 

Below is higher level overview of the issues being fixed:
* #160 (atomic renames aren't working with relative paths)
* File can get corrupted upon exception in both atomic writing functions
* Durability for existing files is broken, since copying is done with `copyFile` and we cannot guarantee successful fsync when handle is closed
* Temporary files are not deleted upon an exception
* `withBinaryFileDurable` produces an error with `ReadMode`
* Possible race condition in between `doesFileExist` and `copyFile`, which can cause an unexpected `DoesNotExist` exception
* Permissions of files are not copied over during atomic rename
* Temporary file is visible on the file system while it is being modified, as such it can be tempered with by another thread or even a different process. By using `open` with `O_TMPFILE` we can provide isolation, which is the implemented approach in this PR.

Besides addition of `writeBinaryFileAtomic` and `withBinaryFileAtomic`, this PR also includes a bunch of tests for the all functions in `RIO.Files` (close to 100% coverage).

Few general changes:
* moved `-DWINDOWS` flag into the `library` section, all tests are OS agnostic and do not directly depend on `unix` nor `Win32`
* turned on `-Wall` by default.
* removed of all compile time warnings
* Added of `readBinaryFile`, `writeBinaryFile` and `withBinaryFileLazy` as synonyms for consistent naming
* Re-exported `RIO.Prelude.Types` from `RIO.Prelude` (it was imported qualified, which prevented from types being re-exported). `import RIO` is not affected.
* Silenced one hlint warning
* Updated changelog and documentation for `RIO.File`

---
This change is [Reviewable](https://reviewable.io/reviews/commercialhaskell/rio/167)

### DIFF
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -9,6 +9,9 @@
 * Addition of `RIO.NonEmpty` module
 * Addition of `RIO.NonEmpty.Partial` module
 * Export `NonEmpty` type from RIO.Prelude.Types
+* Addition of `writeBinaryFileAtomic` and `withBinaryFileAtomic`.
+* Addition of `readBinaryFile`, `writeBinaryFile` and `withBinaryFileLazy`.
+* Fix for [#160](https://github.com/commercialhaskell/rio/issues/160) and a few other atomic+durable related bugs.
 
 ## 0.1.9.2
 
--- a/rio/cbits/rio.c
+++ b/rio/cbits/rio.c
@@ -1,0 +1,34 @@
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+
+int __rio_o_tmpfile( void )
+{
+#ifdef __O_TMPFILE
+  return __O_TMPFILE;
+#else
+  return 0;
+#endif
+}
+
+int __rio_at_fdcwd( void )
+{
+  return AT_FDCWD;
+}
+
+int __rio_at_symlink_follow( void )
+{
+  return AT_SYMLINK_FOLLOW;
+}
+
+
+int __rio_s_irusr( void )
+{
+  return S_IRUSR;
+}
+
+int __rio_s_iwusr( void )
+{
+  return S_IWUSR;
+}
+
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -34,15 +34,12 @@ dependencies:
 # Just for preprocessor directives
 - process
 
+ghc-options:
+- -Wall
+
 when:
 - condition: os(windows)
-  then:
-    cpp-options: -DWINDOWS
-    dependencies:
-    - Win32
-  else:
-    dependencies:
-    - unix
+  cpp-options: -DWINDOWS
 
 library:
   source-dirs: src/
@@ -109,6 +106,16 @@ library:
     - RIO.Prelude.Trace
     - RIO.Prelude.URef
 
+  when:
+  - condition: os(windows)
+    then:
+      dependencies:
+      - Win32
+    else:
+      dependencies:
+      - unix
+      c-sources: cbits/rio.c
+      cc-options: -Wall
 tests:
   spec:
     source-dirs: test
--- a/rio/src/RIO.hs
+++ b/rio/src/RIO.hs
@@ -91,7 +91,6 @@ import RIO.Deque
 import RIO.Prelude
 import RIO.Prelude.Display
 import RIO.Prelude.Exit
-import RIO.Prelude.Extra
 import RIO.Prelude.IO
 import RIO.Prelude.Lens
 import RIO.Prelude.Logger
@@ -99,7 +98,6 @@ import RIO.Prelude.Renames
 import RIO.Prelude.RIO as MonadRIO (RIO(..), liftRIO, runRIO)
 import RIO.Prelude.RIO as SomeRef hiding (RIO(..), liftRIO, runRIO)
 import RIO.Prelude.Simple
-import RIO.Prelude.Text
 import RIO.Prelude.Trace
 import RIO.Prelude.Types
 import RIO.Prelude.URef
--- a/rio/src/RIO/Deque.hs
+++ b/rio/src/RIO/Deque.hs
@@ -274,7 +274,7 @@ dequeToVector :: (VG.Vector v' a, V.MVector v a, PrimMonad m)
 dequeToVector dq = do
     size <- getDequeSize dq
     mv <- V.unsafeNew size
-    foldlDeque (\i e -> V.unsafeWrite mv i e >> pure (i+1)) 0 dq
+    _ <- foldlDeque (\i e -> V.unsafeWrite mv i e >> pure (i+1)) 0 dq
     VG.unsafeFreeze mv
 
 
@@ -317,7 +317,7 @@ freezeDeque ::
   => Deque (VG.Mutable v) (PrimState m) a
   -> m (v a)
 freezeDeque (Deque var) = do
-    state@(DequeState v _ size) <- readMutVar var
+    state@(DequeState _ _ size) <- readMutVar var
     v' <- V.unsafeNew size
     makeCopy v' state
     VG.unsafeFreeze v'
--- a/rio/src/RIO/File.hs
+++ b/rio/src/RIO/File.hs
@@ -174,8 +174,8 @@ foreign import ccall safe "fcntl.h renameat"
   c_safe_renameat :: CInt -> CFilePath -> CInt -> CFilePath -> IO CInt
 
 c_renameat :: DirFd -> CFilePath -> DirFd -> CFilePath -> IO CInt
-c_renameat (DirFd (Fd fdFrom)) cFpFrom (DirFd (Fd fdTo)) cFdTo =
-  c_safe_renameat fdFrom cFpFrom fdTo cFdTo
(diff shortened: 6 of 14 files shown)


lehins:
___
*[rio/src/RIO/File.hs, line 228 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdiQ5DrC21CQVnxSF6Q:-LjMmFdNBzQOB86WctQJ:b-hhhdae) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/src/RIO/File.hs#L228)):*
I see what you mean. I honestly don't like that suggestion.

IMHO the type signature and functionality of this function is strictly worse:
```
openFileFromDir :: MonadIO m => DirFd -> FilePath -> FilePath -> IOMode -> m Handle
openFileFromDir dirFd relPathFromDir fullPathForErrors iomode =
```
than this:
```
openFileFromDir :: MonadIO m => DirFd -> FilePath -> IOMode -> m Handle
openFileFromDir dirFd filePath@(takeFileName -> fileName) iomode
```
With the former it is way to easy to mix up the arguments and get into unpredictable behavior.  There is still an invariant that the file must exist in the folder that `DirFd` is pointing to, thus we are not gaining anything there. It is much easier to describe that invariant in the documentation of `openFileFromDir`

One more important point is that it allows for anybody to think that this is correct, which it is not, because in that case the full path will be used and the `dirFd` will simply be ignored:
```
withDirectory "/home/alexey/.aws" $ \dirFd ->
        withFileInDirectory dirFd "/home/alexey/.aws/credentials" "/home/alexey/.aws/credentials"
```

See `man openat`:
> If pathname is absolute, then dirfd is ignored.

Moreover, I think looking at the call site it is pretty clear that we are dealing with the same file:
```
      withDirectory (takeDirectory filePath) $ \dirFd ->
        withFileInDirectory dirFd filePath
```

One more point is, which I am not quite sure about, but I think I'am correct in my assumption is that doing something like that:
```
withDirectory "/home/alexey" $ \dirFd ->
        withFileInDirectory dirFd ".aws/credentials" "/home/alexey/.aws/credentials"
```
would mess up durable writes, since than it would `fsync` the `dirFd`(`/home/alexey`), but not the `/home/lehins/.aws`, which is the containing folder for the `credentials` file.

lehins:
___
*[rio/src/RIO/File.hs, line 331 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdihWrEAdfXnHlbepGd:-LjMto_7-sVYzCI5jLui:b-5tkw73) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/src/RIO/File.hs#L331)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

I see. I'm happy with that explanation, but still want to to point out:

> most of that information is a lot easier to understand from just following the code rather than trying to understand my comments

I still don't buy that: I understand it now, but it's because of your answer to my question.

As you pointed out: There are many cross-function invariants in this 800-lines file which are neither enforced nor stated. It means one cannot judge the correctness of a given function in isolation. Figuring out all invariants from "just following the code" and keeping it all in working memory is _hard_.  It is much easier to understand your comments.
</blockquote></details>

It has 440 lines of code, it is 800 lines file only because of comments. :P
I'll tell you more, I just read what I said above 2 moths ago and it didn't really make sense to me. I then looked at where `filePath` was used and it was clear to me how it is used.

lehins:
___
*[rio/src/RIO/File.hs, line 178 at r3](https://reviewable.io/reviews/commercialhaskell/rio/167#-LjLSOFJ8PvUtm7Yv-p-:-LjMpE0U8Yx_CtiGvK3q:b-tm0a5i) ([raw file](https://github.com/commercialhaskell/rio/blob/53e53f5109c41de64504d6aadc84274aca65c48a/rio/src/RIO/File.hs#L178)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

Is `cFdTo` a typo of `cFpTo` (it's a `CFilePath`, not an `Fd`)?
</blockquote></details>

Yes, it was a typo. Thanks.

lehins:
___
*[rio/src/RIO/File.hs, line 304 at r3](https://reviewable.io/reviews/commercialhaskell/rio/167#-LjLeXqT37NqZIwpYeNK:-LjMrRJS6z1zpLmtPyw7:b-bfkcq5) ([raw file](https://github.com/commercialhaskell/rio/blob/53e53f5109c41de64504d6aadc84274aca65c48a/rio/src/RIO/File.hs#L304)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

`than` -> `then`

There seems to be some word missing in the sentence, I can't quite parse it:

> If the file descriptor for the directory the target file is/will be located in

Maybe something like "if given"?
</blockquote></details>

I just tried to read that piece of documentation myself and it didn't make any sense to me either. :D
All I had to do was look at the code and I saw what I was trying to say:
```
     -- ^ If a file descriptor is given for the directory where the target file is/will be
     -- located in, then it will be used for opening an anonymous file. Otherwise
     -- anonymous will be opened unattached to any file path.
```
See, it is a lot easier to make human language ambiguous, at least for me ;)

Fixed.

lehins:
___
*[rio/src/RIO/File.hs, line 407 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdjDTna-vJ12nlAvbu4:-LjMy7ga4zTUkDQTzK50:b-ehfyp0) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/src/RIO/File.hs#L407)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

What I mean is that the user-facing documentation (e.g. module or exported-function haddocks) should explain how those temporary files are named.

Here is an example of how this information is important, from a production server breakage system that occurred to me 3 days ago:

I had an `rsync` loop to backup files from one server to another. Software that wrote into the directory was upgraded to do atomic renames, like we add here. Because of the way rsync works (first obtaining file names, then doing the copy), it is possible that it sees a `.xxx.tmp` file that's gone when it wants to copy it, and thus rsync errored out with `file vanished` errors, and I got paged. The correct solution here is to not backup temporary files, using the `--exclude` flag. For that you have to know how the temporary files are named.
</blockquote></details>

Totally agree with you, if that function was exported, except it is not. So comment here would really be useless for the user and for the developer it is easier to see it in the code, rather than read comments.

At the same time, there is no mention of this temporary file naming pattern anywhere in the functions that are exported, I'll add the naming pattern to the relevant places, which will be included in generated haddock.

lehins:
___
*[rio/src/RIO/File.hs, line 453 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdjG6RT3PanERYvOIVs:-LjN-_668QdnO31ATLdE:b4tpqbk) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/src/RIO/File.hs#L453)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

I am not sure what the best name for it should be. But for example in one callee, `withBinaryTempFileFor`, this becomes a `target file` instead of a `source file`.

Doesn't `source ` suggest that we read from that file somehow? As far as I can tell, we don't read from it.
</blockquote></details>

I'll simply remove the comment for this argument. You can't name it, I don't want to name it. This function is not exported, so I really don't care.

lehins:
___
*[rio/src/RIO/File.hs, line 598 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdjKXIN9KU3ff97Qroj:-LjN1G_-3VzXcT93yiBB:b-896fix) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/src/RIO/File.hs#L598)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

Yes
</blockquote></details>

Done.

lehins:
___
*[rio/test/RIO/FileSpec.hs, line 121 at r1](https://reviewable.io/reviews/commercialhaskell/rio/167#-LdjTA9MBlDGCB1Av2i2:-LjN53-uCRphly4GKCm0:b-3frcr0) ([raw file](https://github.com/commercialhaskell/rio/blob/f3f61ad72603976286b7e87bef3da00c0ac9ecc8/rio/test/RIO/FileSpec.hs#L121)):*
<details><summary><i>Previously, nh2 (Niklas Hambüchen) wrote…</i></summary><blockquote>

There's another instance of this just above
</blockquote></details>

Force of habit relying on generators. Fixed.

lehins:
This should be closed in favor of fpco/unliftio#48 Once it is merged in unliftio I'll create a separate PR with simple re-exports.

lehins:
Modified version of the code from this PR was merged into `unliftio`: fpco/unliftio#48
Closing in favor of #196

Issue #169, opened 2019-04-25
Export Data.Ord.Down from RIO


tejon:
Currently the only thing not exported from that module. Don't see much reason to withhold it. It's useful, standard across multiple preludes, and the only conflicts I see via Hoogle relate to keypress handling.

snoyberg:
Addressed by #170

PR #170, opened 2019-04-26
Re-export Data.Ord.Down from RIO.Prelude

tejon:
Addresses #169 

The only reservation I have here is that for some reason there's no record for deconstructing a `Down` (e.g. `getDown`, `unDown`). But there also doesn't seem to be any competing implementation which does provide that.


snoyberg:
I'm not a huge fan of the name, but it is standard, so may as well include it. Thanks!

Issue #168, opened 2019-04-24
RIO.NonEmpty ?


tejon:
`Data.List.NonEmpty` needs a qualified import. Might be worth moving it under `RIO` to standardize the recommended qualification (and encourage its use, maybe).

snoyberg:
Addressed by #173

tejon:
Reopening because #173 failed to export `NonEmpty` from `RIO.Prelude`, which it probably should.

tejon:
Addressed by #173 and #174.

PR #174, opened 2019-04-29
Export Data.List.NonEmpty.NonEmpty from RIO.Prelude.Types

tejon:
I've chosen not to export the `(:|)` constructor here due to conflicts in `generic-random` and `opaleye`. Generally the smart de/constructor functions in `RIO.NonEmpty` are better anyway.


tejon:
Addendum to #168

PR #179, opened 2019-05-14
Fix handling of non-ASCII characters in logSticky

### LINKED ISSUE #178: logSticky is broken for non-ASCII messages
Take the following example:

```haskell
import RIO
import qualified RIO.Text

main = do
    runSimpleApp $ do
        logInfo "one"
        logSticky $ display $ RIO.Text.replicate 41 "ö"
        logInfo "two"
        logSticky "sticky"
```

At the end of the run, you would expect your terminal to look like this:

```
one
two
sticky
```

Yet what actually happens on an 80-width terminal is this:

```
one
two                                    sticky
```

If you change the umlaut `ö` in the `logSticky` to an `o` everything works as expected. Likewise, if you replicate the `ö` a number of times that is less than or equal to half the width of your terminal.

The reason for this weirdness are the functions to clear the sticky line used in `stickyImpl` in [RIO.Prelude.Logger](https://github.com/commercialhaskell/rio/blob/6bceed97b30a799941907c9ae5c7779cae1d2a77/rio/src/RIO/Prelude/Logger.hs#L587):

```haskell
  let backSpaceChar = '\8'
      repeating = mconcat . replicate (B.length sticky) . char7
      clear = logSend lo
        (repeating backSpaceChar <>
        repeating ' ' <>
        repeating backSpaceChar)
```

The variable `sticky` here is a `ByteString` that holds the current sticky line. By using `B.length` to determine how many backspaces are needed to clear it, we overcount if there are non-ASCII characters in the string that are encoded with multiple bytes (like `ö` under UTF-8 which encodes to `\195\182`).

I'm not sure what the best fix is here:

1. We could calculate the number of characters (as opposed to the number of bytes) using something like `Data.ByteString.UTF8.length` from the `utf8-string` package. This is a potentially expensive operation I'm guessing, but if we store the count together with the sticky line, we would only need to do it once when the line is logged (as opposed to every time we need to clear it). Also, the sticky line is going to be maximum as long as the terminal, which is probably <100 characters in most cases.

2. We could clear the sticky line in some other way, e.g. using the functions from `ansi-terminal`. I don't know how portable those are, but this might be the cleanest solution.

3. We could leave it as-is and add a warning to the documentation of `logSticky` to caution against using it with non-ASCII strings.

mcschroeder:
Fixes #178

I’ve also added a test case.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE BangPatterns #-}
 module RIO.Prelude.Logger
   ( -- ** Running with logging
     withLogFunc
@@ -60,6 +61,7 @@ import Control.Monad.IO.Class (MonadIO, liftIO)
 import GHC.Stack (HasCallStack, CallStack, SrcLoc (..), getCallStack, callStack)
 import Data.Time
 import qualified Data.Text.IO as TIO
+import Data.Bits
 import Data.ByteString.Builder (toLazyByteString, char7, byteString, hPutBuilder)
 import Data.ByteString.Builder.Extra (flush)
 import           GHC.IO.Handle.Internals         (wantWritableHandle)
@@ -359,12 +361,12 @@ getCanUseUnicode = do
 newLogFunc :: (MonadIO n, MonadIO m) => LogOptions -> n (LogFunc, m ())
 newLogFunc options =
   if logTerminal options then do
-    var <- newMVar mempty
+    var <- newMVar (mempty,0)
     return (LogFunc
              { unLogFunc = stickyImpl var options (simpleLogFunc options)
              , lfOptions = Just options
              }
-           , do state <- takeMVar var
+           , do (state,_) <- takeMVar var
                 unless (B.null state) (liftIO $ logSend options "\n")
            )
   else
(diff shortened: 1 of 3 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -596,19 +597,32 @@ stickyImpl ref lo logFunc loc src level msgOrig = modifyMVar_ ref $ \sticky -> d
...
-          return sticky
-      | otherwise -> return sticky
+          return (sticky,stickyLen)
+      | otherwise -> return (sticky,stickyLen)
+
+-- | The number of Unicode characters in a UTF-8 encoded byte string.
+utf8CharacterCount :: ByteString -> Int
+utf8CharacterCount = go 0
+  where
+    go n bs = case B.uncons bs of
```
snoyberg: I think the `n` should have a bang pattern on it to avoid accumulating thunks.

mcschroeder: 🤦‍♂ Of course, silly me..

mcschroeder:
Don’t merge this yet, I’ve found another problem: ANSI escape codes. If the sticky log line contains an ANSI escape code, e.g. to color its output, those will also be over-deleted by the `clear` function. Escape codes should not count as characters.

I guess technically all non-printable characters should be ignored when counting the length for clearing purposes. But in practice most of those would screw up the line anyway, except for the color codes. So I think just detecting ANSI color escape sequences and ignoring them when counting would be enough. Then we could have full-color sticky log lines! 🌈 

I’ll see if I can come up with a reasonably simple solution to this.

mcschroeder:
OK, I think this is ready. It now also handles ANSI CSI sequences correctly.

PR #189, opened 2019-06-22
Expanding Microlens

fosskers:
These extended exports should be enough to ensure that most users won't need to depend on `microlens` independently.

Note that I've removed the custom reimplementation of `view`, and instead pulled it from `microlens-mtl`, the transitive deps of which were already in `rio`.

### DIFF
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -1,5 +1,5 @@
 name: rio
-version: 0.1.15.1
+version: 0.1.16.0
 synopsis: A standard library for Haskell
 description: See README and Haddocks at <https://www.stackage.org/package/rio>
 license: MIT
@@ -63,6 +63,7 @@ library:
     - RIO.HashMap
     - RIO.HashMap.Partial
     - RIO.HashSet
+    - RIO.Lens
     - RIO.List
     - RIO.List.Partial
     - RIO.Map
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -7,6 +7,7 @@ author: Michael Snoyman
 maintainer: michael@snoyman.com
 github: commercialhaskell/rio
 category: Control
+
 extra-source-files:
 - README.md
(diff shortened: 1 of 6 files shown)


snoyberg:
I'd feel a bit more comfortable including this as an explicitly imported module like `RIO.Lens` instead of including in the main module itself. Some of the names can be confusing, and some (like `to`) are more likely to cause a name conflict.

fosskers:
Good point - that should be a fairly easy change. Note that this would induce a major version change, this the existing operators exposed by `RIO` would be moved into `RIO.Lens`.

fosskers:
There's precedent for having a `RIO.*` module which doesn't provide anything to the top-level `RIO`. `RIO.Process` is such a module - if you want anything from it, you need to add the extra import line. Is the idea that Lens is specific / specialized enough that we'd want none of it exposed in the top-level `RIO`?

snoyberg:
I disagree. I think the most common lens stuff which doesn't conflict naming-wise can be exported from `RIO`, with the stuff less commonly used or with potential name conflicts exported by `RIO.Lens`.

fosskers:
Ah, gotcha. I'll put up a new commit that proposes a reorg.

fosskers:
@snoyberg There, I've added the `microlens` extras to a new `RIO.Lens` module, and the existing `RIO.Prelude.Lens` exports have been extended a bit.

rio/package.yaml
```diff
@@ -7,6 +7,7 @@ author: Michael Snoyman
 maintainer: michael@snoyman.com
 github: commercialhaskell/rio
 category: Control
+
```
snoyberg, line 10: Can you also include a minor version bump here, a changelog entry, and a `@since` comment on the new module's Haddocks?

fosskers:
That should do it.

PR #192, opened 2019-06-26
exporting 'empty' for 'class Applicative f => Alternative f'

ekalosak:
Prelude wasn't exporting `empty` for `class Alternative f` so this one-line change adds that to the `Prelude.hs` re-export of the `Control.Applicative` module.

### DIFF
--- a/rio/src/RIO/Prelude.hs
+++ b/rio/src/RIO/Prelude.hs
@@ -256,6 +256,7 @@ module RIO.Prelude
     -- * @Alternative@
     -- | Re-exported from "Control.Applicative":
   , (Control.Applicative.<|>)
+  , Control.Applicative.empty
   , Control.Applicative.some
   , Control.Applicative.many
   , Control.Applicative.optional


snoyberg:
We've avoided the export till now because empty is such a common identifier
in the ecosystem and exposing it from a prelude can cause conflicts.

ekalosak:
That makes sense - I wonder if then it's not worth changing `Alternative`'s definition? Are there other creative solutions or should we expect an `import Control.Applicative (empty)` when it's needed?

tejon:
Changing the definition is a non-starter, we're not going to break compatibility with `base`.

Adding an alternate binding e.g. `aempty = Alternative.empty` and exposing _that_ is an option; and though in general we try not to introduce original idioms (preferring established standards), there's precedent in stuff like `fromStrictBytes`. A quick stackage-hoogle search turns up no existing `aempty` so I'm tentatively in favor of going this way. @snoyberg ?

snoyberg:
Exporting `aempty` seems fine. Keep in mind that this will work for _using_ `empty`, but it you want to write a new typeclass instance you'll still have to deal with extra imports.

tejon:
Good point -- IMO that's fine, but let's make sure the haddocks for `aempty` are clear that it's an alias.

rio/src/RIO/Prelude.hs
```diff
@@ -443,3 +444,6 @@ import qualified Data.Text.Encoding.Error (lenientDecode)
 
 import qualified Control.Monad.Primitive (primitive)
 import qualified Control.Monad.ST
+
+aempty :: Control.Applicative.Alternative f => f a
```
snoyberg: Please include a minor version bump, a Haddock comment explaining what this identifier is, a `@since` comment, and a ChangeLog entry.

ekalosak: For sure - commit incoming.
Is there a preferred code formatter?

ekalosak:
(the force pushes were to reword commit messages, nothing else.)

rio/ChangeLog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for rio
 
+## 0.1.10.1
```
snoyberg: This would need to be a minor version bump to 0.1.11.0.

rio/src/RIO/Prelude.hs
```diff
@@ -256,6 +256,9 @@ module RIO.Prelude
     -- * @Alternative@
     -- | Re-exported from "Control.Applicative":
   , (Control.Applicative.<|>)
+  -- | The Alternative class's "empty" function re-exported as "aempty" to avoid
```
snoyberg: The docs need to go below, where the identifier is defined.

rio/src/RIO/Prelude.hs
snoyberg: No, please try to match the existing style, but no specific code formatter.

snoyberg:
The Haddocks are still incorrect, please see the build log. It would be best if you tried to build the docs locally

PR #198, opened 2019-08-24
Adds logFormat to LogOptions

### LINKED ISSUE #197: Be able to add extra context information to logging
It would be good to be able to add extra project specific information to log functions, to add context information automatically. For example, in a current project I'm working on, I have several threads open to manage network connections and it would be good to add which address a log message is referring to.  Currently a lot of the methods to access these parts of a `LogFunc` are private, making it hard to customise as much as I'd like in a project.

I think this could be implemented by adding a method to format the `Utf8Builder` by either:
* Transforming `unLogFunc` to format the message sent.  This would just change the message that's sent to `logGeneric`. This does mean to use with `withLogFunc` you'd need to transform it before embedding it in the monad.
* Add an optional method to `LogOptions` that formats the `Utf8Builder`, which `logGeneric` could call on the message first, before calling the actual log function. Could be `logFormat :: !(Maybe (Utf8Builder -> Utf8Builder))`

Let me know if you think any of these are a good idea (or if I'm missing some other way of going about this!) and I'll have a go at getting a PR up.

RaoulHC:
Closes #197

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -19,6 +19,7 @@ module RIO.Prelude.Logger
   , setLogUseTime
   , setLogUseColor
   , setLogUseLoc
+  , setLogFormat
     -- ** Standard logging functions
   , logDebug
   , logInfo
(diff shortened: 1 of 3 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -509,7 +521,7 @@ simpleLogFunc lo cs _src level msg = do
         timestamp <>
         getLevel logVerbose <>
         ansi reset <>
-        msg <>
+        (logFormat lo) msg <>
```
snoyberg: Minor point, but the parens aren't needed here.

snoyberg:
Thanks!

PR #201, opened 2019-09-22
Move slice to Partial, add sliceMaybe function

### LINKED ISSUE #200: Move `slice` functions from vector to *.Partial
The `slice` functions from vector are partial, which I assumed means they should be moved to the `*.Partial` modules? Applies to:

- RIO.Vector.Unboxed
- RIO.Vector.Storable
- RIO.Vector.Boxed
- RIO.Vector

Perhaps add safe versions returning `Maybe (Vector a)`?

bjornbm:
Closes #200.

### DIFF
--- a/rio/src/RIO/Vector.hs
+++ b/rio/src/RIO/Vector.hs
@@ -20,7 +20,6 @@ module RIO.Vector
   , (Data.Vector.Generic.!?)
 
   -- ** Extracting subvectors
-  , Data.Vector.Generic.slice
   , Data.Vector.Generic.take
   , Data.Vector.Generic.drop
   , Data.Vector.Generic.splitAt
--- a/rio/src/RIO/Vector.hs
+++ b/rio/src/RIO/Vector.hs
@@ -20,6 +20,7 @@ module RIO.Vector
   , (Data.Vector.Generic.!?)
 
   -- ** Extracting subvectors
+  , sliceMaybe
   , Data.Vector.Generic.take
   , Data.Vector.Generic.drop
   , Data.Vector.Generic.splitAt
@@ -256,3 +257,20 @@ module RIO.Vector
   ) where
 
 import qualified Data.Vector.Generic
+import Data.Vector.Generic (Vector)
+import Control.Monad (guard)
+
+-- | /O(1)/ Yield a slice of the vector without copying it. If the vector
+-- cannot satisfy the specificed slice 'Nothing' is returned.
+--
+-- @since 0.1.13.0
+sliceMaybe :: Vector v a
+  => Int  -- ^ @i@ starting index
+  -> Int  -- ^ @n@ length
+  -> v a
+  -> Maybe (v a)
+sliceMaybe i n v = do
+  guard $ i >= 0
+  guard $ n >= 0
+  guard $ Data.Vector.Generic.length v >= i + n
+  pure $ Data.Vector.Generic.slice i n v
--- a/rio/src/RIO/Vector/Partial.hs
+++ b/rio/src/RIO/Vector/Partial.hs
@@ -17,6 +17,7 @@ module RIO.Vector.Partial
   -- ** Extracting subvectors
   , Data.Vector.Generic.init
   , Data.Vector.Generic.tail
+  , Data.Vector.Generic.slice
 
   -- * Modifying vectors
   -- ** Bulk updates
--- a/rio/test/RIO/VectorSpec.hs
+++ b/rio/test/RIO/VectorSpec.hs
@@ -1,0 +1,28 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+module RIO.VectorSpec where
+
+import Test.Hspec
+import RIO
+import qualified RIO.Vector.Boxed as VB
+
+spec :: Spec
+spec =
+  describe "sliceMaybe" $ do
+    it "fails on negative index" $
+      VB.sliceMaybe (-1) 0 v `shouldBe` Nothing
+    it "fails on negative slice length" $
+      VB.sliceMaybe 0 (-1) v `shouldBe` Nothing
+    it "fails when index too large" $
+      VB.sliceMaybe (VB.length v + 1) 0 v `shouldBe` Nothing
+    it "fails when slice too large" $
+      VB.sliceMaybe 0 (VB.length v + 1) v `shouldBe` Nothing
+    it "works for all of vector" $
+      VB.sliceMaybe 0 (VB.length v) v `shouldBe` Just v
+    it "works for zero slice length" $
+      VB.sliceMaybe 0 0 v `shouldBe` Just mempty
+    it "works for zero vector" $
+      VB.sliceMaybe 0 0 (mempty :: VB.Vector ()) `shouldBe` Just mempty
+    it "works for end of vector" $
+      VB.sliceMaybe (VB.length v) 0 v `shouldBe` Just mempty
+  where
+    v = VB.fromList "foobar"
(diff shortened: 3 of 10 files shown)


rio/test/RIO/VectorSpec.hs
```diff
@@ -0,0 +1,28 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+module RIO.VectorSpec where
+
+import Test.Hspec
+import RIO
+import qualified RIO.Vector.Boxed as VB
+
+spec :: Spec
+spec =
```
snoyberg: Some quickchecks/property tests would be a great addition here.

rio/src/RIO/Vector.hs
```diff
@@ -257,3 +257,20 @@ module RIO.Vector
...
+-- @since 0.1.13.0
+sliceMaybe :: Vector v a
+  => Int  -- ^ @i@ starting index
+  -> Int  -- ^ @n@ length
+  -> v a
+  -> Maybe (v a)
+sliceMaybe i n v = do
+  guard $ i >= 0
+  guard $ n >= 0
+  guard $ Data.Vector.Generic.length v >= i + n
```
snoyberg: I'm fairly certain this is subject to an integer overflow error: if `i` and `n` are both very large, they'll be positive, their sum will overflow and can be negative, and then it will be less than `length v`.

bjornbm: Good point. What is the canonical way to check for this?
1.  `i + n >= i`
2.  `n <= maxBound - i`
3.  …?

rio/test/RIO/VectorSpec.hs
bjornbm: Ah, I didn't write any because I didn't see any in `ListSpec.hs`.

rio/src/RIO/Vector.hs
snoyberg: I'm fairly certain you can replace your current check with this:

```haskell
guard $ Data.Vector.Generic.length v - n >= i
```

We know that `length v`, `n` and `i` are all non-negative, and there are more negative `Int`s than positive (due to design of 2's complement). Therefore, `length v - n` is guaranteed not to overflow. That means that, if `n` is too large, `length v - n` is guaranteed to be negative, which is guaranteed to fail the test.

bjornbm: Agreed. Thanks

rio/test/RIO/VectorSpec.hs
bjornbm: I was about to replace all the tests with: 

```hs
spec :: Spec
spec =
  describe "sliceMaybe" $ do
    prop "is consistent with `slice` (pathological cases)" $
      \(QC.Large i) (QC.Large n) v
        -> sliceProp i n (DV.fromList v)
    prop "is consistent with `slice` (less pathological cases)" $
      \(QC.NonNegative i) (QC.NonNegative n) (QC.NonEmpty v)
        -> sliceProp i n (DV.fromList v)
  where
    sliceProp :: Int -> Int -> DV.Vector Char -> QC.Property
    sliceProp i n v = QC.withMaxSuccess 10000 $ case VB.sliceMaybe i n v of
      Just v' -> DV.slice i n v `shouldBe` v'
      Nothing -> evaluate (DV.slice i n v) `shouldThrow` anyException
```

(The second property could be skipped since it is implicit in the first and the number of cases could be reduced to the default of 100, but the coverage may not be extensive.)

However, these tests resulted in:

```
rio        > RIO.Vector       
rio        >   sliceMaybe     
rio        >     is consistent with `slice` (pathological cases) FAILED [1]
rio        >     is consistent with `slice` (less pathological cases)
rio        >       +++ OK, passed 10000 tests.
rio        >                  
rio        > Failures:        
rio        >                  
rio        >   test/RIO/VectorSpec.hs:23:18: 
rio        >   1) RIO.Vector.sliceMaybe is consistent with `slice` (pathological cases)
rio        >        Falsifiable (after 82 tests and 25 shrinks):
rio        >          Large {getLarge = 2104973872246037021}
rio        >          Large {getLarge = 7118398164608738787}
rio        >          ""      
rio        >        did not get expected exception: SomeException
```

Further investigation led me to open <https://gitlab.haskell.org/ghc/ghc/issues/17233>.

bjornbm:
Take a look at 0e59957 and 8673eb7 see if you like any of the approaches.

rio/test/RIO/VectorSpec.hs
snoyberg: What makes you think this is a GHC bug instead of a `vector` library bug?

rio/src/RIO/Vector/Partial.hs
```diff
@@ -62,3 +63,23 @@ module RIO.Vector.Partial
...
 import qualified Data.Vector.Generic
+
+-- | /O(1)/ Yield a slice of the vector without copying it. The vector must
+-- contain at least @i+n@ elements.
+slice :: Data.Vector.Generic.Vector v a
+  => Int   -- ^ @i@ starting index
+  -> Int   -- ^ @n@ length
+  -> v a
+  -> v a
+slice i n v = if i > 0 && n > 0 && i + n < 0  -- `i+n` overflows
```
snoyberg: I'd rather see a patch like this land in `vector` than here. I'm also concerned about surprising behavior of error types changing.

rio/test/RIO/VectorSpec.hs
```diff
@@ -0,0 +1,27 @@
...
+import Test.Hspec
+import Test.Hspec.QuickCheck (prop)
+import qualified Test.QuickCheck as QC
+import RIO
+import qualified RIO.Vector as V
+import qualified RIO.Vector.Partial as V'
+
+spec :: Spec
+spec =
+  describe "sliceMaybe" $ do
```
snoyberg: I don't think it's necessary to _replace_ the unit tests with property tests, I was just hoping to include property tests as well.

rio/test/RIO/VectorSpec.hs
bjornbm: I just used the heuristic that if it is a segfault or the error message says Please report this as a GHC bug it should be reported here. But it may well be (probably is?) a vector bug so I reported it as https://github.com/haskell/vector/issues/257 too.

rio/test/RIO/VectorSpec.hs
bjornbm: That was understood. However, I felt the property test was more to the point and completely specified the intended behavior of `sliceMaybe` (in terms of `slice`), thereby making the unit tests redundant.

Things became a bit complicated with the revealed bug of course.

rio/src/RIO/Vector/Partial.hs
bjornbm: OK, I will remove 8673eb7c7f07530ae9108ec0490f5e277320bc12 then.

rio/src/RIO/Vector.hs
```diff
@@ -257,3 +257,20 @@ module RIO.Vector
...
+sliceMaybe :: Vector v a
+  => Int  -- ^ @i@ starting index
+  -> Int  -- ^ @n@ length
+  -> v a
+  -> Maybe (v a)
+sliceMaybe i n v = do
+  guard $ i >= 0
+  guard $ n >= 0
+  guard $ Data.Vector.Generic.length v - n >= i
+  pure $ Data.Vector.Generic.slice i n v
```
snoyberg: Since we're doing the checking ourselves, moving to `unsafeSlice` is a good idea.

lehins:
`slice` has finally been fixed in `vector-0.12.1`  haskell/vector#257

PR #204, opened 2020-01-01
Introduce withLazyFileUtf8

ony:
We have `withLazyFile`/`readFileBinary`/`writeFileBinary`. I think `readFileUtf8` and `writeFileUtf8` are missing a convenient third friend...

### DIFF
--- a/rio/ChangeLog.md
+++ b/rio/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for rio
 
+## 0.1.13.0
+
+* Add `withLazyFileUtf8`.
+
 ## 0.1.12.0
 
 * Add `logFormat` and `setLogFormat` for `LogOptions`.
--- a/rio/package.yaml
+++ b/rio/package.yaml
@@ -1,5 +1,5 @@
 name: rio
-version: 0.1.12.0
+version: 0.1.13.0
 synopsis: A standard library for Haskell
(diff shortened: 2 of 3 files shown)


ony:
As per [`Data.Text.Lazy.IO` performance notes](http://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-Lazy-IO.html#g:1) it is not clear if it is good to have it in prelude of RIO. I think it should be conscious choice. Most likely that was a reason to not provide this function from beginning.

Though [`Data.Text.IO` have similar notes for performance](http://hackage.haskell.org/package/text-1.2.4.0/docs/Data-Text-IO.html#g:1). So now I'm not sure if it is good idea to have `readFileUtf8` and `writeFileUtf8`... Maybe we need to have more clear documentation about this.

snoyberg:
Overall seems reasonable to me. I don't know that there's any basis for the performance comments in those docs, though there may be. The fact that this code explicitly sets the character encoding likely negates some of the comments over there.

Before that can be merged, please add a changelog entry and version bump. The `@since` comment is already correct.

ony:
@snoyberg, done. Though I'm not sure if this small change worth of dedicated release.

I assumed that small changes being accumulated until the one that makes significant difference for consumers and then they all together being released.

PR #206, opened 2020-01-07
Fix #205 On Windows, findExecutable uses ProcessContext consistently

### LINKED ISSUE #205: Windows: RIO.Process.findExecutable ignores PATHEXT and is inconsistent
On Windows, `RIO.Process.mkProcessContext`, and consequently `RIO.Process.mkDefaultProcessContext`, make use of Windows' environment variable `PATHEXT`, or default to the list `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`, to set the process context 'executable' extensions because:

````haskell
pcExeExtensions =
            if isWindows
                then let pathext = fromMaybe
                           ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC"
                           (Map.lookup "PATHEXT" tm)
                      in map T.unpack $ "" : T.splitOn ";" pathext
                else [""]
````
However, `RIO.Process.findExecutable`, on the `findExecutable name =` limb only - not the `findExecutable name0 | any FP.isPathSeparator name0 =` limb, filters out all extensions other than `.bat", ".cmd", ".com", ".exe` (or upper case equivalents):

````haskell
findExecutable
  :: (MonadIO m, MonadReader env m, HasProcessContext env)
  => String            -- ^ Name of executable
  -> m (Either ProcessException FilePath) -- ^ Full path to that executable on success
findExecutable name0 | any FP.isPathSeparator name0 = do
    pc <- view processContextL
    let names0 = map (name0 ++) (pcExeExtensions pc)
        testNames [] = return $ Left $ ExecutableNotFoundAt name0
        testNames (name:names) = do
            exists <- liftIO $ D.doesFileExist name
            if exists
                then do
                    path <- liftIO $ D.canonicalizePath name
                    return $ return path
                else testNames names
    testNames names0
findExecutable name = do
    pc <- view processContextL
    m <- readIORef $ pcExeCache pc
    epath <- case Map.lookup name m of
        Just epath -> return epath
        Nothing -> do
            let loop [] = return $ Left $ ExecutableNotFound name (pcPath pc)
                loop (dir:dirs) = do
                    let fp0 = dir FP.</> name
                        fps0 = map (fp0 ++) (pcExeExtensions pc)
                        testFPs [] = loop dirs
                        testFPs (fp:fps) = do
                            exists <- D.doesFileExist fp
                            existsExec <- if exists then liftM D.executable $ D.getPermissions fp else return False
                            if existsExec
                                then do
                                    fp' <- D.makeAbsolute fp
                                    return $ return fp'
                                else testFPs fps
                    testFPs fps0
            epath <- liftIO $ loop $ pcPath pc
            () <- atomicModifyIORef (pcExeCache pc) $ \m' ->
                (Map.insert name epath m', ())
            return epath
    return epath
````
because of the `existsExec` test: `D.executable $ D.getPermissions fp` makes use of `System.Directory.getPermissions` and that, for Windows, has (via `getAccessPermissions`):

```` haskell
getAccessPermissions :: FilePath -> IO Permissions
getAccessPermissions path = do
  m <- getFileMetadata path
  let isDir = fileTypeIsDirectory (fileTypeFromMetadata m)
  let w = hasWriteMode (modeFromMetadata m)
  let x = (toLower <$> takeExtension path)
          `elem` [".bat", ".cmd", ".com", ".exe"]
  pure Permissions
       { readable   = True
       , writable   = w
       , executable = x && not isDir
       , searchable = isDir
       }
````

A little knowledge is a dangerous thing, but I think `RIO.Process.findExecutable` should:

(1) also respect the `pcExeExtensions` in the `ProcessContext`; and

(2) take a consistent approach to `if existsExec` (applying it or not) in both of its limbs. Currently, `findExecutable "app\\myProg"` (first limb) behaves differently to `findExecutable "myProg"` (second limb) if `myProg` is an executable with no extension to its file name.

### PR DESCRIPTION
**Pull request by mpilgrem**
The `findExecutable name =` limb of `RIO.Process.findExecutable` no longer filters by the `System.Directory.executable` test and, so, takes into account fully the list of executable extensions in the `ProcessContext`.

This also extends the Haddock documentation for `ProcessContext`, `mkProcessContext`, `mkDefaultProcessContext` and `findExecutable`.

On Windows, an excutable with a filename with no extension cannot be executed from the command line. `mkProcessContext` and `findExecutable` are modified accordingly. `mkProcessContext` now excludes `""` from the head of the list of extensions and the limbs of `findExecutable` use a `isWindows && FP.hasExtension` test to decide whether to include the name itself in the list of names tried.

The `isWindows` function is moved from a `where` to be a generally-available helper function.

### DIFF
--- a/rio/src/RIO/Process.hs
+++ b/rio/src/RIO/Process.hs
@@ -72,6 +72,7 @@ module RIO.Process
     -- * Utilities
   , doesExecutableExist
   , findExecutable
+  , exeExtensions
   , augmentPath
   , augmentPathMap
   , showProcessArgDebug
@@ -225,6 +226,13 @@ currentEnvVarFormat =
   EVFNotWindows
 #endif
 
+-- Don't use CPP so that the Windows code path is at least type checked
+-- regularly
+isWindows :: Bool
+isWindows = case currentEnvVarFormat of
+              EVFWindows -> True
+              EVFNotWindows -> False
+
 -- | Override the working directory processes run in. @Nothing@ means
 -- the current process's working directory.
 --
@@ -271,10 +279,9 @@ mkProcessContext tm' = do
         , pcExeCache = ref
         , pcExeExtensions =
             if isWindows
-                then let pathext = fromMaybe
-                           ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC"
-                           (Map.lookup "PATHEXT" tm)
-                      in map T.unpack $ "" : T.splitOn ";" pathext
+                then let pathext = fromMaybe defaultPATHEXT
+                                             (Map.lookup "PATHEXT" tm)
+                      in map T.unpack $ T.splitOn ";" pathext
                 else [""]
         , pcWorkingDir = Nothing
         }
@@ -283,13 +290,11 @@ mkProcessContext tm' = do
     tm
         | isWindows = Map.fromList $ map (first T.toUpper) $ Map.toList tm'
         | otherwise = tm'
-
-    -- Don't use CPP so that the Windows code path is at least type checked
-    -- regularly
-    isWindows =
-        case currentEnvVarFormat of
-            EVFWindows -> True
-            EVFNotWindows -> False
+    -- Default value for PATHTEXT on Windows versions after Windows XP. (The
+    -- documentation of the default at
+    -- https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start
+    -- is incomplete.)
+    defaultPATHEXT = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC"
 
 -- | Reset the executable cache.
 --
@@ -299,7 +304,10 @@ resetExeCache = do
   pc <- view processContextL
   atomicModifyIORef (pcExeCache pc) (const mempty)
 
--- | Load up an 'EnvOverride' from the standard environment.
+-- | Same as 'mkProcessContext' but uses the system environment (from
+-- 'System.Environment.getEnvironment').
+--
+-- @since 0.0.3.0
 mkDefaultProcessContext :: MonadIO m => m ProcessContext
 mkDefaultProcessContext =
     liftIO $
@@ -307,10 +315,8 @@ mkDefaultProcessContext =
           mkProcessContext
         . Map.fromList . map (T.pack *** T.pack)
 
--- | Modify the environment variables of a 'ProcessContext'.
---
--- This will keep other settings unchanged, in particular the working
--- directory.
+-- | Modify the environment variables of a 'ProcessContext'. This will not
+-- change the working directory.
 --
 -- Note that this requires 'MonadIO', as it will create a new 'IORef'
 -- for the cache.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rio/src/RIO/Process.hs [hunk 1]
```diff
@@ -157,6 +157,9 @@ type EnvVars = Map Text Text
 
 -- | Context in which to run processes.
 --
+-- On Windows, the context includes filename extensions associated with
```
snoyberg: I don't think this comment adds a lot: for non-Windows users it's confusing, and for Windows users it doesn't explain what this means. I'd recommend such a comment be added to any exported identifier allowing the user to override these.

rio/src/RIO/Process.hs
```diff
@@ -258,6 +268,10 @@ exeSearchPathL = processContextL.to pcPath
 
 -- | Create a new 'ProcessContext' from the given environment variable map.
 --
+-- On Windows, the filename extensions associated with executables are from the
```
snoyberg: Again, this should be related to an identifier which allows these to be overridden.

rio/src/RIO/Process.hs
```diff
@@ -582,12 +602,15 @@ findExecutable name = do
...
-                        fps0 = map (fp0 ++) (pcExeExtensions pc)
+                               -- On Windows, include the name if it has an
+                               -- existing extension
+                        fps0 = (if isWindows && FP.hasExtension fp0
+                                  then (fp0:) else id)
+                               (map (fp0 ++) (pcExeExtensions pc))
                         testFPs [] = loop dirs
                         testFPs (fp:fps) = do
                             exists <- D.doesFileExist fp
-                            existsExec <- if exists then liftM D.executable $ D.getPermissions fp else return False
```
snoyberg: This looks broken: it isn't checking for the executable bit on non-Windows platforms.

rio/src/RIO/Process.hs
mpilgrem: OK, I have deleted it. There is no exported identifier allowing the user to override a `ProcessContext`, in this regard. I'll explain what I have done further below.

rio/src/RIO/Process.hs
mpilgrem: I have made this much shorter and added an 'only' to 'On Windows only, ...' to reinforce that non-Windows users can skip it. As there is no exported identifier, I have kept the documentation with this most general of the 'smart' constructors - and avoided the repetition with the more specialised 'smart' constructor `mkDefaultProcessContext` by referring back to `mkProcessContext`. I think some Haddock documentation is necessary because behaving as if the system `PATHEXT` variable is defined with a default value, when it is not present in the environment, is not something you might anticipate.

rio/src/RIO/Process.hs
mpilgrem: You are correct. In the original 'non-Windows' code there was an inconsistency in the two limbs. My version had no `isExecutable` filter in both limbs. I've now added the `isExecutable` filter to both limbs in a revised version. I need, however, to do some testing - it will be a little while until I have access to a non-Windows machine to do that.

mpilgrem:
I have made changes to this request (and updated the commit message to reflect them) but I need to do some testing on Windows and non-Windows machines. I will comment further when I have completed that.

The review comments about an exported identifier allowing the user to override the relevant part of a `ProcessContext` made me realise that it would be helpful, at least, to export an `exeExtensions` utility that was the `ProcessContext` equivalent of `System.Directory.exeExtension`. Users would have a way to know what extensions, part of the `ProcessContext`, were affecting the behaviour of `findExecutable`.

mpilgrem:
I have tested successfully on Windows and non-Windows (macOS). The details of my testing are set out below.

__On Windows (version 10.0.18362.535):__

I confirmed the behaviour of commands at the Windows' Command Prompt, as follows:

```
>REM Create a small sample application, myProg.exe
>echo main = putStrLn "Hello World!" > myProg.hs
>stack exec ghc -- myProg.hs
[1 of 1] Compiling Main             ( myProg.hs, myProg.o )
Linking myProg.exe ...

>myProg
Hello World!

>rename myProg.exe myProg
>myProg
'myProg' is not recognized as an internal or external command,
operable program or batch file.

>rename myProg myProg.app
>myProg.app
Hello World!
```

The same behaviour occurs if a path is given (_eg_ `>mydir\myProg.app`) or if the file is on the `PATH`.

I tested `findExecutable` with a simple application, `rioTest.exe` based on:

```haskell
module Main (main) where

import System.Environment (getArgs)

import RIO
import RIO.Process

main :: IO ()
main = test =<< (head <$> getArgs)

test :: String -> IO ()
test name = runSimpleApp $ do
  result <- findExecutable name
  liftIO $ case result of
    Left err -> print err
    Right p -> print p
```
I tested the following combinations of file names/locations and arguments to `findExecutable`:

|File                   |rioTest _arg_   |Result   |Pass?|
|-----------------------|----------------|---------|:---:|
|myProg.exe             |myProg          |_Located_|Y    |
|myProg                 |myProg          |Executable named myProg not found on path:|Y|
|myProg.app             |myProg.app      |_Located_|Y    |
|mydir\myProg.exe       |mydir\myProg    |_Located_|Y    |
|mydir\myProg           |mydir\myProg    |Did not find executable at specified path:|Y|
|mydir\myProg.app       |mydir\myProg.app|_Located_|Y    |
|`PATH`\myProg.exe      |myProg          |_Located_|Y    |
|`PATH`\myProg.exe      |myProg.exe      |_Located_|Y    |
|`PATH`\myProg          |myProg          |Executable named myProg not found on path:|Y|
|`PATH`\myProg.app      |myProg.app      |_Located_|Y    |
|`PATH`\mydir\myProg.exe|mydir\myProg    |Did not find executable at specified path:|Y|

__On non-Windows (macOS version 10.13.6)__

I created a small sample executable, as for Windows above (`echo 'main = putStrLn "Hello World!"' > myProg.hs`) and a non-executable file `test` (`touch test`) which had the same name as an executable on the `PATH`; and tested with the same `rioTest` executable, as for Windows above.

On macOS, `.` is not on the `PATH` in Terminal by default and, for non-Windows, the (original and current) `mkProcessContext` does not add `"."` to the head of `pcPath`. So, I added `.` to the head of the `PATH` (`PATH=.:$PATH`).

I tested the following combinations of file names/locations and arguments to `findExecutable`:

|File                      |rioTest _arg_   |Result   |Pass?|
|--------------------------|----------------|---------|:---:|
|myProg (`.` not on `PATH`)|myProg          |Executable named myProg not found on path:|Y|
|myProg (`.` now on `PATH`)    |myProg          |_Located_|Y    |
|myProg.exe                |myProg          |Executable named myProg not found on path:|Y|
|myProg.exe                |myProg.exe      |_Located_|Y    |
|test                      |test            |_Non-execuatable ignored and `test` command elsewhere on the path located_|Y|
|myProg.hs                 |myProg.hs       |Executable named myProg.hs not found on path:|Y|
|mydir/myProg              |mydir/myProg    |_Located_|Y    |
|mydir/myProg.exe          |mydir/myProg.exe|_Located_|Y    |
|mydir/myProg.exe          |mydir/myProg    |Did not find executable at specified path: mydir/myProg|Y|
|mydir/test                |mydir/test      |Did not find executable at specified path: mdir/test|Y|
|`PATH`/myProg (other than `./myProg` already tested above)            |myProg          |_Located_|Y    |

rio/src/RIO/Process.hs
```diff
@@ -258,6 +266,10 @@ exeSearchPathL = processContextL.to pcPath
 
 -- | Create a new 'ProcessContext' from the given environment variable map.
 --
+-- On Windows only, if the system @PATHEXT@ environment variable is absent from
```
snoyberg: This isn't really true. This implies that a new `PATHEXT` value is added to the env vars. My original concern still also stands: I don't believe this function is the right place for this explanation, it should go exclusively on an identifier which is related to the path extensions.

rio/src/RIO/Process.hs
```diff
@@ -563,11 +582,16 @@ findExecutable
...
-    let names0 = map (name0 ++) (pcExeExtensions pc)
+                 -- On Windows, include the name if it has an existing extension
+    let names0 = (if isWindows && FP.hasExtension name0 then (name0:) else id)
+                 (map (name0 ++) (pcExeExtensions pc))
         testNames [] = return $ Left $ ExecutableNotFoundAt name0
         testNames (name:names) = do
             exists <- liftIO $ D.doesFileExist name
-            if exists
+            existsExec <- if exists
+                              then liftIO $ (isWindows ||) <$> isExecutable name
```
snoyberg: I'd rather see `if isWindows then pure True else liftIO $ isExecutable name`. It's clearer in purpose IMO, and doesn't get in the way of shortcircuiting behavior.

rio/src/RIO/Process.hs
```diff
@@ -582,11 +606,17 @@ findExecutable name = do
             let loop [] = return $ Left $ ExecutableNotFound name (pcPath pc)
                 loop (dir:dirs) = do
                     let fp0 = dir FP.</> name
-                        fps0 = map (fp0 ++) (pcExeExtensions pc)
+                               -- On Windows, include the name if it has an
+                               -- existing extension
+                        fps0 = (if isWindows && FP.hasExtension fp0
```
snoyberg: Any way to share this logic and the `isExecutable` logic from above?

rio/src/RIO/Process.hs
```diff
@@ -599,6 +629,22 @@ findExecutable name = do
             return epath
     return epath
 
+-- | Test whether a file has executable permissions.
+isExecutable :: FilePath -> IO Bool
```
snoyberg: I think best would be to include a `Bool` argument here indicating if it's Windows or not.

rio/src/RIO/Process.hs
mpilgrem: Which other exported identifier are you thinking of? The constructors of type `ProcessContext` are not exported and this behaviour is introduced by the `mkProcessContext` 'smart' constructor function. There is no function that manipulates the `pcPath` field other than via `mkProcessContext`.

(As an aside, I don't understand why `mkProcessContext` behaves in this way. A Windows user could legitimately - if unusually -  want to remove `PATHEXT` from the environment so as not to have the functionality on Windows that `PATHEXT` facilitates.)

In terms of the documentation itself, instead of:

"On Windows only, if the system `PATHEXT` environment variable is absent from the given map, `mkProcessContext` behaves as if the variable were present with its default value.`

how about:

"On Windows only, if the system `PATHEXT` environment variable is absent from the given map, `mkProcessContext` behaves as if the variable were present with its default value, but without adding it to the environment variables of the process context.`

rio/src/RIO/Process.hs
mpilgrem: Will do.

rio/src/RIO/Process.hs
mpilgrem: There is a way. I will include it in a revised pull request.

rio/src/RIO/Process.hs
mpilgrem: Understood.

mpilgrem:
Subject to the discussion of where/how to document the behaviour of `mkProcessContext` on Windows when `PATHEXT` is absent from the environment, I think the revised pull request reflects the most recent review. In particular, I have refactored `findExecutable` to remove the code duplication (and updated the commit message accordingly). I have reperformed the tests that I describe above, on Windows and non-Windows.

rio/src/RIO/Process.hs
snoyberg: I wrote up a response to this which seems to have gotten lost, sorry.

I think this comment is approaching this backwards. It's saying we pretend like there's a field present, except for this one thing. Except the one thing is the only functionality (setting env vars) that people think it should be relevant to. It leaves open a bunch of questions to the reader, _especially_ the reader on a non-Windows machine, because they don't have the context of this problem in their mind.

Instead: I'd rather see the functions that actually look up executables say directly: on Windows, we use PATHEXT, and if it isn't there, we fall back to a default. I'd rather have that documentation in one place, ideally on the new `exeExtensions` accessor. And then I'd like to see it referenced from the relevant places. Right now, the concepts are spread across too many pieces of documentation, especially where it isn't relevant to most people.

mpilgrem: Thank you. I hope my revised pull request now reflects that.

rio/src/RIO/Process.hs
```diff
@@ -309,8 +316,8 @@ mkDefaultProcessContext =
 
 -- | Modify the environment variables of a 'ProcessContext'.
 --
--- This will keep other settings unchanged, in particular the working
--- directory.
+-- This will keep other settings that do not depend on the environment variables
```
snoyberg: I don't understand what this change is adding. The new language sounds more confusing to me. Is this relevant to this PR?

rio/src/RIO/Process.hs
```diff
@@ -299,7 +304,9 @@ resetExeCache = do
   pc <- view processContextL
   atomicModifyIORef (pcExeCache pc) (const mempty)
 
--- | Load up an 'EnvOverride' from the standard environment.
+-- | As for 'mkProcessContext' but specialised to the system environment (from
```
snoyberg:
```suggestion
-- | Same as 'mkProcessContext' but uses the system environment (from
```

Issue #213, opened 2020-03-25
documentation: it's not clear what the "source" argument to logInfoS and friends is for


robx:
While trying to figure out a nice way to provide context to log messages (e.g., prepend some request-specific prefix in a web handler), I ran across [`logInfoS`](https://hackage.haskell.org/package/rio-0.1.14.1/docs/RIO.html#v:logInfoS). From the documentation, it was entirely unclear to me what this "source" might be. It'd be great to be a bit clearer about this.

(I've since read the source code and think that it's actually never used by log functions created through `newLogFunc`, so perhaps these might even be dropped? I'm going to assume I'm missing something though.)

snoyberg:
PR to clear up the docs would be great. Source is intended to be a general purpose field for "what part of your application does this come from." For example, you may say source is "database" to distinguish from "web."

robx:
Thanks for looking into this! Would be happy to provide such a PR, but am still a bit confused.

Is it correct that in order to get the source argument displayed, I would have to create my own log function via `mkLogFunction`, and not be able to build on top of `withLogFunc` / `newLogFunc`? That's how I read the source code...

snoyberg:
I think so.

PR #215, opened 2020-04-01
Document log source

robx:
This is to address #213.

I fear it rather shows up the limits of the current API rather than provide a really useful example, but I suppose that's helpful for the reader anyway.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -35,6 +35,8 @@ module RIO.Prelude.Logger
   , logSticky
   , logStickyDone
     -- *** With source
+    --
+    -- $withSource
   , logDebugS
   , logInfoS
   , logWarnS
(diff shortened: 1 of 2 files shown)


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -204,6 +206,37 @@ logOther
   -> m ()
 logOther = logGeneric "" . LevelOther
 
+-- $withSource
+--
+-- There is a set of logging functions that take an extra 'LogSource'
+-- argument to provide context, typically detailing what part of an
+-- application the message comes from.
+--
+-- The built-in log functions as constructed with 'withLogFunc' and
```
snoyberg, line 215: Actually, maybe the right thing to do instead is simply to start including the source when present. I don't think that in any way a meaningful breaking change. Thoughts? And would you be interested in amending this PR to include that?

robx: Sure! Will give it a shot.

robx:
Closing in favour of #216.

PR #219, opened 2020-05-12
Re-export Hashable typeclass functions

locallycompact:


snoyberg:
I think we left these out because they tended to cause a lot of naming conflicts. But we included `Hashable` itself because it was useful in adding constraints. Since using the methods directly is so uncommon, I'd lean towards exposing these from outside of the `RIO` module itself, such as in `RIO.Hashable`.

PR #222, opened 2020-07-31
Add colours to the `LogOption` constructor

mpilgrem:
The motivation for this pull request is that the exisitng hard-coded colours in `RIO.Process.withProcessTimeLog` and `RIO.Prelude.Logger.simpleLogFunc` (for timestamps and locs) are not visible when using the Solarised Dark theme.

Tested on Windows 10 by sucessfully rebuilding `stack` (which depends on `rio`), with and without making use of the option to customise logger colours.

The defaults are the same as what was previously hard-coded and `RIO.Prelude.Logger` does not export the constructor of `LogOption`, so existing code depending on `rio` should not be affected.

A new `LogColors` type is exported, and an associated accessor (`logFuncColorsL`):

```haskell
data LogColors = LogColors
  { -- | The color associated with each 'LogLevel'.
    logColorLogLevels :: !(LogLevel -> Utf8Builder)
    -- | The color of secondary content.
  , logColorSecondary :: !Utf8Builder
    -- | The color of accents, which are indexed by 'Int'.
  , logColorAccents :: !(Int -> Utf8Builder)
  }
```

timestamps and locs in `simpleLogFunc` use the colour of secondary content. `withProcessTimeLog` uses accent colour 0 to highlight the duration time.

### DIFF
--- a/rio/src/RIO/Prelude/Logger.hs
+++ b/rio/src/RIO/Prelude/Logger.hs
@@ -24,6 +24,9 @@ module RIO.Prelude.Logger
   , setLogUseColor
   , setLogUseLoc
   , setLogFormat
+  , setLogLevelColors
+  , setLogSecondaryColor
+  , setLogAccentColors
     -- ** Standard logging functions
   , logDebug
   , logInfo
@@ -56,6 +59,9 @@ module RIO.Prelude.Logger
   , noLogging
     -- ** Accessors
   , logFuncUseColorL
+  , logFuncLogLevelColorsL
+  , logFuncSecondaryColorL
+  , logFuncAccentColorsL
     -- * Type-generic logger
     -- $type-generic-intro
   , glog
@@ -319,6 +325,7 @@ logOptionsMemory = do
         , logTerminal = True
         , logUseTime = False
         , logUseColor = False
+        , logColors = defaultLogColors
         , logUseLoc = False
         , logFormat = id
         , logSend = \new -> atomicModifyIORef' ref $ \old -> (old <> new, ())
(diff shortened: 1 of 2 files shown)


mpilgrem:
Since my original pull request, I have changed the type in the three `setLog...` functions from `Text` to `Utf8Builder`. Originally, I reasoned that `Text` was a more user-friendly type for those functions but now I think that it is wrong to impose the `Text -> Utf8Builder` overhead on all users.

rio/src/RIO/Prelude/Logger.hs
```diff
@@ -459,11 +466,42 @@ data LogOptions = LogOptions
...
   , logUseLoc :: !Bool
   , logFormat :: !(Utf8Builder -> Utf8Builder)
   , logSend :: !(Builder -> IO ())
   }
 
+-- | ANSI color codes for use in the configuration of the creation of a
+-- 'LogFunc'.
+--
+-- @since 0.1.18.0
+data LogColors = LogColors
```
snoyberg, line 479: It seems surprising to me to export the constructors here, not export the `defaultLogColors` value, but yes export the setters. I'd lean towards not exporting the constructor and fields.

mpilgrem: I have revised the pull request. Rather than export `LogColors` and its three constructors, I export only three accessors.

snoyberg: I think you'll also need to export `defaultLogColors`, otherwise it won't be possible to create a value to modify

mpilgrem: I think it is not correct to export identifier `defaultLogColors`, for the folowing reason: with one exception, the only way to create a `LogOptions` value (other than from an existing `LogOptions` value and using the `set...` functions) is to use `RIO.logOptionsHandle` (which is what returns the 'default' `LogOptions` value that is the starting point for others). (The exception is `RIO.logOptionsMemory` which is 'primarily intended for testing' - and also returns a 'default'/'empty' `LogOptions` value.) The 'defaults' are documented but they are 'behind the scenes'.

PR #225, opened 2020-09-14
Expose `fromLeft` and `fromRight` from `RIO.Prelude`

lehins:
These guys have been available since base-4.10 which is already the lower bound for rio.


snoyberg:
They're partial functions, they shouldn't be included

snoyberg:
Doh, I'm an idiot, ignore me, it's early :p

snoyberg (approved):

PR #229, opened 2021-03-19
A (draft of a) mixin-based template

dequbed:
Cabal 2.0/GHC 8.2 (and later) support mixins to rename, show, and hide modules.

We can use those mixins to not have to use `NoImplicitPrelude` and instead treat RIO as a 'true' Prelude-replacement, i.e. changing the import Path to `Prelude`-prefixed so instead of `import RIO.File` one uses `import Prelude.File`.

While this may not be the best way of using RIO it is basically how I've been using RIO in my private projects and I wanted to show this way in case anybody else finds it interesting.
The actual code itself could also be improved a lot but I kept as close as possible to the original template this was copied from to make the differences obvious.
Biggest drawback(s) of this approach is that the file `src/Prelude.hs` which provides the `Prelude` module is absolutely required for the build process, that this fact is not very obvious, and that if you drop this file or have a very different folder structure the errors you will get are very much not helpful.

This template may not be something you *want* in the upstream repo in which case feel free to close this PR and/or give me a pointer to where this would be more suited. Otherwise this still needs to actually be integrated because I don't know how stack templates work and I didn't want to break anybodies builds by making a mistake there.

### DIFF
--- a/etc/mixin-template/.gitignore
+++ b/etc/mixin-template/.gitignore
@@ -1,0 +1,4 @@
+*~
+*.swp
+tarballs/
+.stack-work/
--- a/etc/mixin-template/ChangeLog.md
+++ b/etc/mixin-template/ChangeLog.md
@@ -1,0 +1,3 @@
+# Changelog for PROJECTNAME
+
+## Unreleased changes
--- a/etc/mixin-template/LICENSE
+++ b/etc/mixin-template/LICENSE
@@ -1,0 +1,30 @@
+Copyright AUTHOR (c) 2019
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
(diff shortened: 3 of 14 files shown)


snoyberg:
I'd be happy to include a link to it in a different repo, but I'm not ready to promote it inside this repo. I haven't used mixins myself and am a bit concerned about some of these features. I'm also not certain I understand why we'd want to encourage people to import `Prelude.File` instead of `RIO.File`. And finally, maintaining this list of module renames while `rio` evolves seems like overhead that we'll forget to do.

dequbed:
> I'd be happy to include a link to it in a different repo, but I'm not ready to promote it inside this repo. I haven't used mixins myself and am a bit concerned about some of these features. I'm also not certain I understand why we'd want to encourage people to import `Prelude.File` instead of `RIO.File`. And finally, maintaining this list of module renames while `rio` evolves seems like overhead that we'll forget to do.

Understandable, especially the last point is something I was concerned about as well and I don't see a good solution either. I'll close this PR.

Cheers and thanks for writing & maintaining RIO, it's a wonderful framework for writing Haskell :)

PR #235, opened 2021-08-08
Fix augment path wrt #234

### LINKED ISSUE #234: augmentPathMap assumes uppercase PATH on windows
On windows, the Path variable may not be uppercase `PATH`, but `Path` and is frequently returned as such by e.g. `getEnvironment` and friends. Using it with `augmentPathMap` may then lead to confusing behavior, where a new key is added and gets ignored.

To make this more robust, on windows, it should:

1. query the existing map for a case-insensitive PATH key
2. use that key to insert the new values

If there are multiple PATH keys, we're screwed. There could be a `normalizePathKeys :: EnvVars -> EnvVars` function that forces it to be upper-case and merges duplicate PATHs?
Sth. like:

```hs
normalizePathKeys  :: EnvVars -> EnvVars
normalizePathKeys = Map.foldrWithKey (\k v m -> case T.toLower k of
                                            "path" -> Map.insert "PATH" v m
                                            _      -> Map.insert k v m) mempty
```

hasufell:
On windows, env vars are generally case-insensitive.
Due to `type EnvVars = Map Text Text` augmentPathMap
assumes full uppercase "PATH" variable, which may lead
to surprising behavior if the current map
has a variable such as "Path".

This patch folds over the map and inserts any path
key as "PATH" on windows. That also means: if there
are multiple, the "last" one wins. There generally
isn't a sane solution if the input map already
has multiple path keys (how should they be merged,
which order?).

Fixes #234

### DIFF
--- a/rio/src/RIO/Process.hs
+++ b/rio/src/RIO/Process.hs
@@ -190,7 +190,7 @@ data ProcessException
     | ExecutableNotFound String [FilePath]
     | ExecutableNotFoundAt FilePath
     | PathsInvalidInPath [FilePath]
-    deriving Typeable
+    deriving (Typeable, Eq)
 instance Show ProcessException where
     show NoPathFound = "PATH not found in ProcessContext"
     show (ExecutableNotFound name path) = concat
@@ -654,7 +654,8 @@ exeExtensions = do
   pc <- view processContextL
   return $ pcExeExtensions pc
 
--- | Augment the PATH environment variable with the given extra paths.
+-- | Augment the PATH environment variable with the given extra paths,
+-- which are prepended (as in: they take precedence).
 --
 -- @since 0.0.3.0
 augmentPath :: [FilePath] -> Maybe Text -> Either ProcessException Text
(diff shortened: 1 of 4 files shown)


rio/src/RIO/Process.hs
```diff
@@ -672,9 +673,21 @@ augmentPath dirs mpath =
...
-     return $ Map.insert "PATH" path origEnv
+     return $ Map.insert "PATH" path normEnv
   where
-    mpath = Map.lookup "PATH" origEnv
+    -- On windows, env vars are case insensitive.
+    -- We pick the "last" PATH variable and ensure
+    -- it's full uppercase.
+    -- See https://github.com/commercialhaskell/rio/issues/234
+    normEnv 
+      | isWindows = Map.foldrWithKey
```
snoyberg, line 683: This is a large change in the complexity of the operation. I'd rather see this addressed at the ingestion period, as I mentioned in the issue.

hasufell: I only see two ways to do this earlier:

1. create a smart constructor for `EnvVars` (breaking API change)
2. fix `getEnvironment` and friends in `base` (I'm not even sure that's a good idea)

For my use case `EnvVars` is constructed like so:

```hs
    curEnv <- Map.fromList . map (T.pack *** T.pack) <$> liftIO getEnvironment
```

Since the type alias is exposed, we can assume similar uses in the wild.

snoyberg:
Now I'm actually a bit more confused. I was looking through the code base to try and understand why this hasn't occurred previously, and it looks like we're already upper casing all keys on creation of a `ProcessContext`:

https://github.com/commercialhaskell/rio/blob/4eba306d37a60a20f0bed27f40afc23a7c892c40/rio/src/RIO/Process.hs#L289-L298

That's what I was talking about with "ingestion," and it seems to already be working. Do you have a minimal repro demonstrating a case where it's _not_ working?

hasufell:
> Do you have a minimal repro demonstrating a case where it's not working?

Yes, the unit test I already added to this PR: https://github.com/commercialhaskell/rio/pull/235/files#diff-3243d978fddb991353ef007ddb5c436c03b772bf44ab35f0e40d31485ca31ad5R24

Basically:

```hs
    let origEnv = [("Path", "/local/bin")]
    mkProcessContext (augmentPathMap ["/bin"] origEnv)
```

The code you showed does this:

```hs
    isWindows = Map.fromList $ map (first T.toUpper) $ Map.toList tm'
```

`Map.fromList` will throw away duplicates:

```
Prelude Data.Map> fromList [("PATH", "/local/bin"), ("PATH", "/bin")]
fromList [("PATH","/bin")]
```

snoyberg:
Now I understand. This isn't the intended workflow when using `rio`. `augmentPathMap` is intended to be used with functions like `modifyEnvVars`. I'd rather go with a doc fix than impact the performance of the library in this way. In your use case, if you want to use `getEnvironment` directly, you can implement the same fix as within the `RIO.Process` module itself, by upper casing the keys on Windows.

PR #237, opened 2021-08-09
Fix augment path

### LINKED ISSUE #234: augmentPathMap assumes uppercase PATH on windows

hasufell:
On windows, env vars are generally case-insensitive.
Due to 'type EnvVars = Map Text Text' augmentPathMap
assumes full uppercase "PATH" variable, which may lead
to surprising behavior if the current map
has a variable such as "Path".

This patch folds over the map and inserts any path
key as "PATH" on windows. That also means: if there
are multiple, the "last" one wins. There generally
isn't a sane solution if the input map already
has multiple path keys (how should they be merged,
which order?).

Fixes #234

### DIFF
--- a/rio/src/RIO/Process.hs
+++ b/rio/src/RIO/Process.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ViewPatterns #-}
 
 -- | Interacting with external processes.
 --
@@ -190,7 +191,7 @@ data ProcessException
     | ExecutableNotFound String [FilePath]
     | ExecutableNotFoundAt FilePath
     | PathsInvalidInPath [FilePath]
-    deriving Typeable
+    deriving (Typeable, Eq)
 instance Show ProcessException where
     show NoPathFound = "PATH not found in ProcessContext"
     show (ExecutableNotFound name path) = concat
@@ -269,7 +270,7 @@ exeSearchPathL = processContextL.to pcPath
 --
 -- @since 0.0.3.0
 mkProcessContext :: MonadIO m => EnvVars -> m ProcessContext
-mkProcessContext tm' = do
+mkProcessContext (normalizePathEnv -> tm) = do
     ref <- newIORef Map.empty
     return ProcessContext
         { pcTextMap = tm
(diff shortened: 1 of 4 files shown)


rio/src/RIO/Process.hs
```diff
@@ -672,9 +673,21 @@ augmentPath dirs mpath =
...
-     return $ Map.insert "PATH" path origEnv
+     return $ Map.insert "PATH" path normEnv
   where
-    mpath = Map.lookup "PATH" origEnv
+    -- On windows, env vars are case insensitive.
+    -- We pick the "last" PATH variable and ensure
+    -- it's full uppercase.
+    -- See https://github.com/commercialhaskell/rio/issues/234
+    normEnv 
+      | isWindows = Map.foldrWithKey
```
hasufell, line 683: If you fancy the other way, it could also be:

```hs
Map.fromList $ map (first T.toUpper) $ Map.toList tm'
```

which is used `mkProcessContext`.

Factor it out?

snoyberg: Yeah, that looks better to me. It's easier to understand, and it _might_ be more efficient (though that would need profiling to analyze, I could be wrong).

Factoring out would be great. thanks!

rio/src/RIO/Process.hs
```diff
@@ -287,17 +288,21 @@ mkProcessContext tm' = do
...
     -- Default value for PATHTEXT on Windows versions after Windows XP. (The
     -- documentation of the default at
     -- https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start
     -- is incomplete.)
     defaultPATHEXT = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC"
 
 
+-- Fix case insensitivity of the PATH environment variable on Windows,
+-- by forcing all keys full uppercase.
+normalizePathEnv :: EnvVars -> EnvVars
```
hasufell, line 300: Not entirely sure if we need an inline pragma here.

snoyberg: I think it's reasonable as-is.

snoyberg (approved):
LGTM. I'll merge/release once CI passes. Thanks!

snoyberg:
Released to Hackage, thanks!

snoyberg:
I'll open a PR against Stack to use this newer version, unless you'd like to open that PR instead.

PR #239, opened 2021-12-02
Fix small typo

fabfianda:
...we will enabled -> we will enable


snoyberg:
Thanks!

fabfianda:
Thank you all for the amazing work you do in the Haskelll world!

PR #242, opened 2022-04-24
Export a generalised version of `augmentPathMap` (as `augmentPathMap'`)

mpilgrem:
Existing `RIO.Process.augmentPathMap` is specialised to the PATH environment variable. There are other environment variables commonly used to contain lists of paths (eg PKG_CONFIG_PATH) and it would be helpful to be able to augment them in the same way. This pull request proposes to extend the API by providing the generalised:

~~~haskell
augmentPathMap' :: Text -> [FilePath] -> EnvVars -> Either ProcessException EnvVars
~~~

Existing `augmentPathMap` then is defined as `augmentPathMap = augmentPathMap' "PATH"`.

`augmentPath` is not specialised to PATH or any other environment variable. This pull request improves its Haddock documentation to clarify that (1) it augments the value of any environment variable assumed to list paths, and (2) it does not check that the existing value lists paths.

### DIFF
--- a/rio/src/RIO/Process.hs
+++ b/rio/src/RIO/Process.hs
@@ -77,6 +77,7 @@ module RIO.Process
   , exeExtensions
   , augmentPath
   , augmentPathMap
+  , augmentPathMap'
   , showProcessArgDebug
     -- * Reexports
   , P.ProcessConfig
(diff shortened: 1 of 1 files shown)


snoyberg (approved):
Thanks!

PR #249, opened 2022-07-28
Typo in compiler, normally hyphenated words.

philderbeast:


mpilgrem:
@philderbeast, thanks!

PR #259, opened 2025-09-14
Accept HLint v3.10 hints for rio package

mpilgrem:
This pull request accepts the HLint v3.10 suggestions. None of them seemed to be controversial. A few are stylistic but not offensive and adopting HLint's style is a path of least resistance.

### DIFF
--- a/rio/src/RIO/Deque.hs
+++ b/rio/src/RIO/Deque.hs
@@ -89,7 +89,7 @@ newDeque
   => m (Deque v (PrimState m) a)
 newDeque = do
     v <- V.new baseSize
-    liftM Deque $ newMutVar (DequeState v 0 0)
+    Deque <$> newMutVar (DequeState v 0 0)
   where
     baseSize = 32
 {-# INLINE newDeque #-}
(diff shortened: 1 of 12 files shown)


akhra:
did you verify that the Core output is identical for the `$!` and lambda substitutions? particularly for the former it's not obvious to me that it should be, and it was clearly done for optimization purposes.

mpilgrem:
@akhra, I think `\_ -> x` => `const x` must be only a question of style?

It also seems to me that `return $! ()` => `return ()` must be uncontroversial.

As for `return $! Just x` => `return (Just x)`, I'll need to think about that if HLint's `f $! x` => `f x` can't be trusted.

akhra:
agreed on your first two; the lambda that caught my attention was `\val -> modifyIORef' ref (\_ -> val)` => `modifyIORef' ref . const` which might be doing something with the _outer_ lambda closure. (I couldn't guess what, but I'm also quite sure Michael knows GHC tricks I don't, so it's just something to verify!)

mpilgrem:
The simplification of `\val -> modifyIORef' ref (\_ -> val)` was suggested in two steps:

1. `\val -> modifyIORef' ref (const val)` -- that seems to me to be uncontroversial
2. `(modifyIORef' ref) . const` -- that step also seemed to me to be uncontroversial

PR #261, opened 2025-09-14
Fix #255 Re-export type operator `~`

### LINKED ISSUE #255: Export `~` of `Data.Type.Equality`?
The `Prelude` and `Data.Type.Equality` of the `base` package export:

~~~haskell
class a ~# b => (a :: k) ~ (b :: k)
~~~

and GHC 9.10.2 (now specified by Stackage LTS Haskell 24.0 snapshot) warns like:

~~~text
 warning: [GHC-12003] [-Wtype-equality-out-of-scope]
    • The ‘~’ operator is out of scope.
      Assuming it to stand for an equality constraint.
    • NB: ‘~’ used to be built-in syntax but now is a regular type operator
          exported from Data.Type.Equality and Prelude.
      If you are using a custom Prelude, consider re-exporting it.
    • This will become an error in a future GHC release.
~~~

See also: https://errors.haskell.org/messages/GHC-12003/

Should it be exported by the `RIO` prelude too?

mpilgrem:
See:
* #255


akhra (approved):
LGTM! and thanks for the ping, my Haskell skills have gotten a bit rusty but small PRs like this are a great way to get back in 🙂

PR #265, opened 2025-10-03
Make `canUseUtf8` use `hGetEncoding`

jeltsch:
There is [an effort to move some GHC-specific modules out of `base`](https://gitlab.haskell.org/ghc/ghc/-/issues/26134). Candidates for such moving are `GHC.IO.Handle.Types` and `GHC.IO.Handle.Internals`. These modules are sometimes used for determining the text encoding that a handle employs. Such a use is also present in the `rio` package as part of the internal operation `RIO.Prelude.Logger.canUseUtf8`.

However, there is no need to resort to the above-mentioned GHC-specific modules for this purpose, as the `System.IO` module contains `hGetEncoding` for determining text encodings of handles. This pull request changes `canUseUtf8` such that it uses `hGetEncoding`.


akhra:
LGTM (and I double-checked that hGetEncoding is available back to base-4.12!) but for some reason CI is persistently failing (three tries) on `macos-latest, stack-ghc-9.10.3`. just that one. @mpilgrem any insight?

mpilgrem:
@akhra, I have solved the CI problem.

akhra:
> Note that `hGetEncoding` internally uses `withHandle_` instead of `wantWritableHandle`, which the previous implementation of `canUseUtf8` uses. It should be checked whether `canUseUtf8` can do without the additional functionality of `wantWritableHandle`, in particular the check that the given handle is not closed.

oop, this got edited in... good catch. `canUseUtf8` is neither documented nor used internally (repository search turns up only the definition) so I'm not 100% sure of the intent, but it's in with the logging system so definitely intended for writable handles. but in any case it shouldn't need the extra read/write state assertion of `wantWritableHandle` — @jeltsch can you incorporate an `hIsWritable` check? that should avoid any potential of a breaking change.

jeltsch:
> `canUseUtf8` is neither documented nor used internally (repository search turns up only the definition) […]

I did find a use: in the same module in the implementation of `logOptionsHandle`. 🙂

I cann add the `hIsWritable` check.

jeltsch:
I just filed [GHC issue #26479](https://gitlab.haskell.org/ghc/ghc/-/issues/26479) about the bug with `hIsWritable` and an analogous bug with `hIsReadable`.

akhra:
saw this, decided to see if the linter cares before asking for another revision... it does, sorry 😛

`when (not isWritable)` -> `unless isWritable`

Issue #269, opened 2025-10-08
Move writability check out of `canUseUtf8` or drop it


jeltsch:
#265 included a writability check to `RIO.Prelude.Logger.canUseUtf8`, because the previous implementation of `canUseUtf8` also had one, although not so explicit, being covered by `wantWritableHandle`. However, this writability check does not belong in `canUseUtf8`, as it has nothing to do with encodings.

We should possibly move the writability check further up so that it is directly done during logger initialization.

An alternative is to simply drop the check and rely on actual writes failing when the handle is not writable. The state of a handle can be changed any time no operation is performed on the handle; in particular, the handle can be closed. Therefore, a check for writability during logger initialization anyhow provides no guarantee that the handle will be writable later, except perhaps if the handle is under the sole control of the logger.

akhra:
addressed via #270, we'll want to move that into CPP once the [upstream fix](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/15227) lands in a release.

PR #270, opened 2025-10-14
remove open/writable check from canUseUtf8, add assertWritable

akhra:
addresses #269


mpilgrem:
@akhra, it looks good to me.

I did wonder if this is an opportunity for the Haddock documentaton for `logOptionsHandle` to note that it throws an illegal operation error if the specified `Handle` is not writeable.

PR #272, opened 2026-01-03
add withLogContext function

AxelUlmestig:
I was using RIO (thanks for an excellent library) and I wanted to add the same extra context to a bunch of log messages in a scope. My first idea was to add this into my own project as a helper function but I couldn't because the LogFunc constructor isn't exposed.

So now I'm trying to add it directly into RIO instead. I hope that you too will find some value in this functionality,  that the documentation snippet I added makes sense and that the code lives up to your standards.

Your feedback is appreciated.

example usage:
```hs

main = do
  runSimpleApp do
    withLogContext ("processing job 123: " <>) do
      logInfo "hello"
```

Prints "processing 123: hello"

### DIFF
--- a/rio/test/RIO/LoggerSpec.hs
+++ b/rio/test/RIO/LoggerSpec.hs
@@ -2,9 +2,9 @@
 {-# LANGUAGE OverloadedStrings #-}
 module RIO.LoggerSpec (spec) where
 
-import Test.Hspec
-import RIO
-import Data.ByteString.Builder (toLazyByteString)
+import           Data.ByteString.Builder (toLazyByteString)
+import           RIO
+import           Test.Hspec
 
 spec :: Spec
 spec = do
@@ -86,3 +86,10 @@ spec = do
       logInfoS "tests" "should appear"
     builder <- readIORef ref
     toLazyByteString builder `shouldBe` "[info] (tests) should appear\n"
+  it "withLogContext" $ do
+    (ref, options) <- logOptionsMemory
+    withLogFunc (options & setLogVerboseFormat True) $ \lf -> runRIO lf $ do
+      withLogContext ("[context] " <>) $
+        logInfoS "tests" "should appear"
+    builder <- readIORef ref
+    toLazyByteString builder `shouldBe` "[info] (tests) [context] should appear\n"


rio/src/RIO/Prelude/Logger.hs
```diff
@@ -817,6 +818,26 @@ logFuncAccentColorsL = logFuncL.to
...
+-- > {-# LANGUAGE OverloadedStrings #-}
+-- >
+-- > main = do
+-- >  runSimpleApp do
+-- >    withLogContext ("processing job 123: " <>) do
+-- >      logInfo "hello"
+--
+-- Prints "processing job 123: hello"
+withLogContext :: (HasLogFunc env, MonadReader env m) => (Utf8Builder -> Utf8Builder) -> m a -> m a
+withLogContext addContext action = local (over logFuncL wrapLogFunc) action
```
akhra: linter wants this eta reduced (i.e. remove `action` from both sides)

rio/test/RIO/LoggerSpec.hs
```diff
@@ -86,3 +86,10 @@ spec = do
       logInfoS "tests" "should appear"
     builder <- readIORef ref
     toLazyByteString builder `shouldBe` "[info] (tests) should appear\n"
+  it "withLogContext" $ do
+    (ref, options) <- logOptionsMemory
+    withLogFunc (options & setLogVerboseFormat True) $ \lf -> runRIO lf $ do
+      withLogContext ("[context] " <>) $
+        logInfoS "tests" "should appear"
+    builder <- readIORef ref
+    toLazyByteString builder `shouldBe` "[info] (tests) [context] should appear\n"
```
akhra: looks technically sound, but I'm not sure this is the right direction. if we want local log context as a supported feature like level and source (and we might!) then it should probably have a similar opinionated/consistent format, rather than being a freeform adjustment like this. which would necessitate a round of bikeshedding that seems entirely unnecessary for what you're actually trying to do here and now.

meanwhile, `addContext` can be `const ""`  -- this is really a total formatting override, not just a context function. and `LogOptions` already has `logFormat :: !(Utf8Builder -> Utf8Builder)` which (at least in `simpleLogFunc`) does exactly the same thing! the snag is, the only place we can currently manipulate it is with `setLogFormat` on the bare LO. it's expected that you're not changing options after the LogFunc is created, which I think is reasonable for all the other fields, but this one was added later and copying the pattern of the rest seems to have been insufficient. (tbh maybe it was a mistake to put it in `LogOptions` instead of `LogFunc`, but that's done.)

let's build from there. I'd suggest two new exported functions (these are untested):
```haskell
-- like setLogFormat, but can incorporate an existing one
modifyLogFormat :: ((Utf8Builder -> Utf8Builder) -> Utf8Builder -> Utf8Builder) -> LogOptions -> LogOptions
modifyLogFormat f' lo = setLogFormat (logFormat options & f') lo

-- like your withLogContext, but explicitly happens *after* extant logFormat
-- (less chance of breaking ANSI codes, etc.)
withLogFormat :: (HasLogFunc env, MonadReader env m) => (Utf8Builder -> Utf8Builder) -> m a -> m a
withLogFormat f = local $ over logFuncL postprocessLogFormat
  where
    postprocessLogFormat (LogFunc lf lo) = LogFunc lf $ modifyLogFormat (f .) lo
```
deliberately not exposing convenient in-environment access to `set`/`modifyLogFormat` because they kinda feel like footguns to me, especially in team settings. I can't think of a use case that isn't either covered here, or really ought to be signposted by building a new LogFunc. happy to revisit that if there are objections/demand, but I think this covers your purpose?

akhra (changes requested):

rio/test/RIO/LoggerSpec.hs
akhra: after some extra thought, how about:
```haskell
-- fully exported
type LogFormat = Utf8Builder -> Utf8Builder
withLogPrepend :: (HasLogFunc env, MonadReader env m) => Utf8Builder -> m a -> m a
withLogAppend :: (HasLogFunc env, MonadReader env m) => Utf8Builder -> m a -> m a

-- exported from Rio.Prelude.Logger but hidden from RIO
withLogFormatMutator :: (HasLogFunc env, MonadReader env m) => (LogFormat -> LogFormat) -> m a -> m a
```
`modifyLogFormat` can downgrade to a helper in `withLogFormatMutator` since it would otherwise only have use before LogOptions construction (where you already have the source functions in scope). and the footgun is available but properly tucked away in the footgun cabinet, slightly unergonomic on purpose so it only comes out when you really know you need it.

AxelUlmestig: The `MonadUnliftIO` constraint is a bit unfortunate, the application that I wanted to use it for was to add some extra information job processing. I was using an `ExceptT` transformer along with RIO there
```hs
type Job env a = Except.ExceptT JobErr (RIO env) a
```
`MonadUnliftIO` only works with `ReaderT` and `IO`, right?

Issue #273, opened 2026-01-18
facility for scoped log formatting


akhra:
hidden constructor/accessors on `LogFunc` and `LogOptions` complicate clean user implementation of local alterations to log output, e.g. adding a scoped prefix or enabling call stack output.

Issue #38, opened 2018-02-03
Defining MonadCatch etc. instances for RIO?


robinp:
I see that there's already a `MonadThrow` instance. Wonder if there's something fundamentally preventing defining `MonadCatch`, `MonadMask` instance for `RIO`? So that  code written against those generic constraints could still be run in `RIO`.

There's a specific usecase I'm interested in where this is not a philosophic difference, that is where one can't just use RIO all the way down: for concurrency testing using [`dejafu`](https://hackage.haskell.org/package/dejafu-1.0.0.1), for which computations need to live in [`MonadConc`](https://hackage.haskell.org/package/concurrency-1.4.0.0/docs/Control-Monad-Conc-Class.html) from `concurrency` package.

Thank you.
+@barrucadu

snoyberg:
I'm on the fence about it. On the one hand: I think encouraging `MonadCatch` and `MonadMask` is a bad idea, as they lead to confusing behavior in many common monads (see https://www.snoyman.com/reveal/monad-transformer-state and https://www.youtube.com/watch?v=KZIN9f9rI34). On the other hand, the `RIO` case is completely safe, and adding the instances adds no extra dependencies. I'm open to input on this.

roman:
I would love to see `concurrency` working nicely with `rio`, is it possible for `concurrency` to stop relying on `MonadCatch` and `MonadMask` and replacing it with whatever `rio` supports (I guess this is `unliftio`)? Also, if `concurrency` does this change, would it stop working with other tools?

I can relate why (even though it doesn't hurt) we would not like the `RIO` monad to extend from `MonadCatch` and `MonadMask`.

barrucadu:
In principle `concurrency` could stop relying on `MonadCatch` and `MonadMask`, but it raises some awkward interface questions.

If `MonadCatch` goes away, what does the type of `forkFinally` become?  Would there need to be two of them: `forkFinally`, which doesn't catch exceptions; and `forkFinallyWithExceptions` which does?  Does `MonadSTM` lose its `MonadCatch` superclass too?  If `MonadMask` goes away, what happens to `forkWithUnmask` and `forkOnWithUnmask`?  Do they get put in a new `MonadConcMask` class?

Or maybe `MonadConc` would remain as it is, but it would get a new superclass `MonadConcWithoutCatchOrMask`, which most of the operations migrate to.

tejon:
Possible short-term solution: put `MonadCatch` and `MonadMask` instances in
[...]

PR #274, opened 2026-01-18
scoped log formatting

akhra:
addresses #273; supersedes #272


akhra:
dug around some more looking for other paths we could take, and reasoning behind where we're at. the good news is I'm pretty confident I've wrapped my head around it. the other news is I've come around to agree with most of the original decisions. :sweat_smile: 

I looked around for a bracketing solution that's got better ergonomics. the only one I could find that looked like it might fit the bill is in the `exceptions` package, and... we're already dependent on `exceptions`, but _only_ for `MonadThrow` (which RIO is an instance of!) and `throwM`. explicitly not for `MonadCatch` or `MonadMask` (required for their `bracket`). well, not within the `rio` package: instead they live in `rio-orphans`. derived instances, nothing special, clearly separated on purpose.

I turned to Google for any kind of explanation and found our very own issue #38, with this comment from Michael:

> I'm on the fence about it. On the one hand: I think encouraging `MonadCatch` and `MonadMask` is a bad idea, as they lead to confusing behavior in many common monads (see https://www.snoyman.com/reveal/monad-transformer-state and https://www.youtube.com/watch?v=KZIN9f9rI34). On the other hand, the `RIO` case is completely safe, and adding the instances adds no extra dependencies. I'm open to input on this.

there was a lot of demand (`concurrently` uses them) and the real salt is that not far down you'll find _**me in 2018**_ (maybe not even a contributor yet?) suggesting they could be put into a quarantine module, pre-deprecated so that it could be dumped as soon as a better solution was found. Michael went with a whole a separate package, and that's why `rio-orphans` exists in the first place.

at that point my thought was, if he was on the fence to begin with and _I'm partly responsible_ for which way the whole situation leaned, could this use case be compelling enough to move those instances over? so I followed the links I quoted above and... no, he was right, this is _absolutely_ not a class of errors I ever want to be fielding issue reports over, to the point that after watching the seminar video I've even discarded the idea of tucking `withLogFunc'` into `rio-orphans`. (I probably watched it in 2018 too, but life has sent me lots of places since then and a refresher was definitely needed.)

---

meanwhile in parallel, I've gone cold again on `composeLogFormat` because it has its own class of frustrating issues. consider:

```haskell
ansiReset   = "<RESET>"   -- "\ESC[0m"
ansiFgCyan  = "<FG-CYAN>" -- "\ESC[36m"
ansiFgBlack = "<FG-BLCK>" -- "\ESC[30m"
ansiBgGray  = "<BG-GRAY>" -- "\ESC[100m"
-- an intended and likely common use of LogFormat
baseFormat msg = ansiFgCyan <> msg <> ansiReset
-- a natural extension of what we're trying to make work here
logPrefix pre msg
  = ansiBgGray <> ansiFgBlack
  <> pre <> ansiReset <> " "
  <> msg

baseLogOptions = setLogFormat baseFormat . setLogUseColor False
withLogPrefix1 pre = withLocalLogFunc $ composeLogFormat ((logPrefix pre) .)
withLogPrefix2 pre = withLocalLogFunc $ composeLogFormat (. (logPrefix pre))

main = runSimpleApp $ withLocalLogFunc baseLogOptions $ do
  logInfo "withLogPrefix1"
  example withLogPrefix1
  logInfo "withLogPrefix2"
  example withLogPrefix2

example wLP = do
  wLP "scoped:" $ do
    logInfo "message"
    wLP "deeper:" $ do
      logInfo "message"
```
(initially tested with actual colors, but this is more readable than the raw escape codes)
```
ghci> main
<FG-CYAN>withLogPrefix1<RESET>
<BG-GRAY><FG-BLCK>scoped:<RESET> <FG-CYAN>message<RESET>
<BG-GRAY><FG-BLCK>deeper:<RESET> <BG-GRAY><FG-BLCK>scoped:<RESET> <FG-CYAN>message<RESET>
<FG-CYAN>withLogPrefix2<RESET>
<FG-CYAN><BG-GRAY><FG-BLCK>scoped:<RESET> message<RESET>
<FG-CYAN><BG-GRAY><FG-BLCK>scoped:<RESET> <BG-GRAY><FG-BLCK>deeper:<RESET> message<RESET>
```
`withLogFormat1` composes with the prefix function on the left. whoops: every time you embed scope the inner prefix comes _before_ the outer one. counterintuitive, let's fix it.

`withLogFormat2` inverts the composition. the prefixes are in the right order! but now their ANSI directives are *inside* the ones from the base formatter, so there's a reset before the intended base color code and a prefixed message comes out terminal default instead of cyan.

with only blind composition, there is no way to solve both problems at once. clearly labeling it as a composition felt like enough to trust that people will generally think through the effects before wielding a power tool... but after enough thinking, it's not really a power tool at all: just a frustrating distraction from cleaner solutions. e.g., storing your own structured scoping info in the environment and wrapping the log functions to consume it. or if you really do just want quick-and-dirty format hacks: explicitly store your base format function in the environment, compose it in your own code, and use `setLogFormat` for updates. 

---

pulling it all together: one of the explicit core goals of RIO is to encourage best practices. if we do something to directly support scoped prefixing, it needs to use a separate and more structured mechanism than just overloading the format function; and that's not a trivial project. hackers gonna hack, power tools and escape hatches are fine, but the stuff we _directly provide_ should be clean of caveats and gotchas.

with `RIO.Orphans` you can roll your own MonadIO-friendly `withLogFunc`. in anything but a `SimpleApp` you'd be calling that yourself anyway: you'll have access to the `LogOptions` before it's encapsulated, you can use all its setters, in fact that's currently the only way it would get a formatting function to begin with. as I suggested above, you can store both that function and your `LogOptions` in an accessible location if you want. ultimately you can in fact hack your way to a bespoke `withLocalLogFunc` in user space as-is.

it has occurred to me that doing so could be less tedious and more smoothly integrated with the addition of `LogFuncOptionsL`. but it's not _necessary_, it does nothing on its own and may cause confusion just by being there, and I'm pretty firm in the reasoning against `composeLogFormat` (and equally `LogFuncLogFormatL`) so you'd need to manually track the function regardless.

a deeply unsatisfying conclusion, that I'm going to sleep on once or twice before closing this PR. maybe the rest of the week will bring fresh ideas, but unfortunately that's where I'm at right now.

Issue #276, opened 2026-01-21
eliminate warnings when compiling RIO


akhra:
from our very own readme,
> For code targeting production use, you should also use the flag that turns all warnings into errors, to force you to resolve the warnings before you ship your code

via `stack build` over `ghc-9.10.3`, we've got several:

- two from -Wx-partial in `RIO.List`
- one from -Wnoncanonical-monoid-instances in `RIO.Prelude.RIO`
- two from -Wdeprecation in `RIO.ByteString`

happily, `cabal build` over `ghc-9.14.1` doesn't add any.

looking into these, I don't anticipate any difficulty. the partiality warnings are on our safe wrappers; there are other ways to write those. the `Monoid` change is trivial: we already require `base >= 4.12` so it can just be removed. likewise, `Bytestring` just wants the same function imported from a different module.

PR #277, opened 2026-01-21
eliminate compiler warnings

akhra:
addresses #276


akhra:
updated to remove additional spurious `mappend` definitions. (for the `Utf8Builder` instance, first verified the `inline` is already present in `base-4.12`.)

mpilgrem:
@akhra, looks good to me.

Issue #278, opened 2026-01-21
remove CPP conditionals supporting base < 4.12


akhra:
`base >= 4.12` is a hard requirement since 0.1.21 (2021).

PR #279, opened 2026-01-21
remove CPP conditional supporting base < 4.12

akhra:
addresses #278 

(I swear *just* I saw more of these, and going back to 4.8. must have been while I was checking old versions to see when we bumped which bounds, to make sure none of the warning fixes needed CPP... well anyway, there was still _one!_)


mpilgrem:
I can't reproduce the CI problem locally on a Mac mini M1, 2020/Tahoe 26.0.1 with:
~~~text
$ stack -w stack-ghc-9.10.3.yaml test
$ stack -w stack-ghc-9.12.2.yaml test
~~~

